PR #2, opened 2017-07-25
Proofread Docs

RealOrangeOne:
- _API_ should be upper case
- Cleanup code examples
- Correct Spelling
- Highlight variables
- General Cleanup


SThomasP:
Looks good

PR #19, opened 2017-10-12
Spelling

RealOrangeOne:


.spelling
```diff
@@ -0,0 +1,11 @@
+Sourcebots
```
PeterJCLaw: Is this one word or two? We seem to have it as one word in some places and two in others. It would be good to decide and record this somewhere.

cc @Adimote.

thomasleese: We decided on SourceBots yesterday.

thomasleese: https://github.com/sourcebots/wiki/wiki/Branding

PeterJCLaw: Aha, cool. #21 updates the spelling file and usages, though I've just spotted that you'd previously already done part of that in #20.

PR #24, opened 2017-10-21
Update 'Python: A whirlwind tour'

RealOrangeOne:
https://trello.com/c/nRqlyApL/1-update-python-a-whirlwind-tour

- Make markdown far nicer
- Update to python3
- Indent code with spaces
- Rearrange Structure


content/tutorials/python-whirlwind-tour.md
```diff
@@ -225,7 +225,7 @@ The code is executed once for each item in the list, with `x` set to each item i
 [ ] Newspaper
 ```
 
-Unfortunately, this method doesn't tell you the index of the current item. `x` is only a temporary variable, so modifying it has no effect on the list itself (try it). This is where the `enumerate` function comes in. It tells us the index of each value we loop over. An example with numbers:
+Unfortunately, this method doesn't tell you the index of the current item. `x` is only a temporary variable, so modifying it has no effect on the list itself (try it). This is where the `enumerate` function comes in (see the [Calling functions](#calling-functions). It tells us the index of each value we loop over. An example with numbers:
```
kierdavis: There seems to be a bracket mismatch here, and the phrase doesn't make sense - are you missing the word "section"?

content/tutorials/python-whirlwind-tour.md
```diff
@@ -335,205 +202,135 @@ Half way there!
...
 my_list = [42, "BOOMERANG!!!", [0, 3]]
-print len(my_list)
-~~~~~
+print(len(my_list))
+```
 
 Output:
 
-~~~~~ not-code
+```python
```
kierdavis: Should this be "text" (or whatever keyword implies no syntax highlighting) instead of "python"?

Whilst the output of a program may be syntactically valid Python, it shouldn't be considered a Python program. IMO presenting source code with highlighting and program output without it gives a clear visual distinction between the two types of verbatim block used in this document.

There's a few other instances of this in the page.

RealOrangeOne: hm, actually yes that does sound like a wise idea, will change

content/tutorials/python-whirlwind-tour.md
RealOrangeOne: it would appear I am

kierdavis:
The markdown looks good to me. I've not tested what the resulting webpage looks like - it's probably a good idea to do so for a change of this size.

RealOrangeOne:
[See webpage](https://user-images.githubusercontent.com/6527489/32138027-74a250e8-bc23-11e7-8038-6268ca7061b9.png)

kierdavis:
Webpage looks fine, thanks.

PR #25, opened 2017-10-21
Servo assembly

RealOrangeOne:
https://trello.com/c/YgyAhMq4/6-write-servo-assembly-documentation


content/api/servo-board/GPIO.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Pin Mode
+GPIO pins support 4 different modes, allowing it to work in 4 different ways. A pin can only have 1 mode at a time, and certain pins aren't compatible with certain modes. These Pin modes are stored in an [enum](https://docs.python.org/3/library/enum.html) which needs to be imported to use.
+
+```python
+from robot import PinMode
+``` 
+
+{{% notice note %}}
+The input modes closely resemble those of an Arduino. More information on them can be found in [their docs](https://docs.python.org/3/library/enum.html)
```
Adimote: wrong link?

content/api/servo-board/GPIO.md
```diff
@@ -0,0 +1,68 @@
...
+{{% /notice %}}
+
+
+### `PinMode.INPUT`
+Allows reading the value from the pin
+
+### `PinMode.INPUT_PULLUP`
+Allows reading the value from the pin using the built-in pullup resistor.
```
Adimote: Feels like we need to explain pullup resistors somewhere, or at least link to a good explanation

content/api/servo-board/GPIO.md
```diff
@@ -0,0 +1,68 @@
...
+```
+
+## Reading a pin value
+
+To read the value of a pin, simply call `read()`. This will return either `PinValue.HIGH` or `PinValue.LOW`, which have to be imported to use.
+
+```python
+from robot import PinMode
+
+r.servo_board.gpio[3].read() == PinValue.HIGH
```
Adimote: Need to import PinValue

content/api/servo-board/GPIO.md
```diff
@@ -0,0 +1,68 @@
...
+>>> True
+```
+
+{{% notice warning %}}
+Calling `read()` will raise an exception if the pins mode hasn't been set to an input type beforehand.
+{{% /notice %}}
+
+# Analogue Pins
+
+Certain sensors output analogue signals rather than digital ones, so have to be read differently. The servo assembly has 4 analogue inputs.
```
Adimote: link to / explain analogue signals?

content/api/servo-board/_index.md
```diff
@@ -0,0 +1,28 @@
...
+```
+
+### Get a specific board
+```python
+board_one = r.servo_boards['S3rv0']
+board_one = r.servo_boards[0]
+```
+
+{{% notice tip %}}
+If you already have a board instance, and want to get its serial number, you can use the `serial` property.
```
Adimote: avoid words like 'instance', I'd suggest rewording this to "if you want to get the serial for a board you can use the `serial` property, i.e. `print(r.servo_boards[0].serial)`

content/api/servo-board/GPIO.md
RealOrangeOne: Agreed, although i'm not entirely sure what one is myself, or at least why it's preferable in some cases. Anyone know a good resource / an provide a short explanation? cc @kierdavis

kierdavis: https://learn.sparkfun.com/tutorials/pull-up-resistors looks like a good resource, and it also talks about internal pullup resistors inside microcontrollers (like we use here).

A disadvantage is that it contains code example(s) for Arduino, which may be confusing since it's a different syntax to our kit. I don't think it's too much of a problem though - especially if we ensure there's an example using our syntax near to this link.

It's not an easy topic to explain succinctly - I feel like any description I could write would still be a few paragraphs long.

content/api/servo-board/GPIO.md
kierdavis: Something brief along the lines of

> Analogue signals can have any voltage, while digital signals can only take on one of two voltages. They need to be connected to one of the servo's assembly's 4 dedicated analogue inputs.

would probably do here.

content/api/servo-board/GPIO.md
```diff
@@ -0,0 +1,73 @@
+---
+title: GPIO
+---
+
+You can use the GPIO pins to read and write individual pins of the servo assembly. These can be used for anything, from microswitches to LEDs. GPIO is only available on pins 2 to 12, because pins 0 and 1 are reserved for communication with our kit.
```
kierdavis: I think it'd be a good idea to state what GPIO stands for here.

content/api/servo-board/GPIO.md
```diff
@@ -0,0 +1,73 @@
...
+
+To read the value of an analogue pin, call `read_analogue()` on the servo board. This will give you the value of all analogue pins.
+
+```python
+r.servo_board.read_analogue()["a0"]
+>>> 1.569
+``` 
+
+{{% notice info %}}
+The values are the voltage read on the pins, between 0 and 5.
```
kierdavis: "voltage" -> "voltages"

content/api/servo-board/servos.md
```diff
@@ -0,0 +1,33 @@
+---
+title: Servos
+---
+
+From the servo assembly, you can control the attached servos. By default, servos will be unpowered when your robot starts, and can freely rotate. Once you set a value, they then fix to that rotation. This also happens when you turn off your robot, or unplug your USB.
```
kierdavis: I'd recommend saying "USB stick" instead of just "USB" (in case it gets misinterpreted as the USB connection to the servo assembly).

content/api/servo-board/servos.md
```diff
@@ -0,0 +1,33 @@
...
+Servos can be controlled using the `position` parameter. This should be set to a value between -1 and 1.  
+
+```python
+servo_one.position = 0.65
+print(servo_one.position)
+>>> 0.65
+```
+
+{{% notice tip %}}
+When `position` is set to 0, the servo loses power and can freely rotate. The same state they're in at start-up.
```
kierdavis: I believe this occurs when `position` is set to `None`, not `0`.

See: https://github.com/sourcebots/robotd/blob/master/robotd/devices.py#L349

content/api/servo-board/ultrasound.md
```diff
@@ -0,0 +1,24 @@
...
+## Reading a distance
+Assuming you wired the _trigger_ pin to pin 6, and the _echo_ pin to pin 7, you can query the distance from the sensor using:
+
+```python
+r.servo_board.read_ultrasound(6, 7)
+>>> 0.524
+```
+
+{{% notice info %}}
+The returned distance is in meters.
```
kierdavis: meters -> metres

kierdavis:
Will wait for @Adimote to take a look before merging.

kierdavis:
What's the reasoning behind that last commit? (just curious)

RealOrangeOne:
@kierdavis related to https://github.com/sourcebots/docs/pull/30

kierdavis:
Ahh I see, cool.

Looks like there's a merge conflict with the `.spelling` file now, could you check that out?

PR #27, opened 2017-10-28
Kit assembly

RealOrangeOne:
Pulled from google drive almost exactly.

Images added as links rather than inline, as floating images around doesnt work well on the web, and i'd have to do it manually!


content/tutorials/kit-assembly.md
```diff
@@ -0,0 +1,57 @@
...
+- Arduino (The blue, exposed board with a metal USB-B connector)
+- Arduino Servo Shield (another blue board, hopefully on top of the first)
+- 1 x 3.81mm Camcon (for the Arduino Servo Shield)
+- 2 x 5mm Camcon (for the Motors)
+- 1 x 5mm Camcon with wire loop (attached to the power board)
+- 2 x 7.5mm Camcon (for the Motor Board)
+- USB Hub
+- USB Webcam
+- 2 x Black USB-A to micro-usb cables
+- Beige USB-A to USB-B cable (for the Arduino)
```
kierdavis: I'd suggest to removing the colours from this line and the one above - we can (and quite possibly will) give them any old colour of USB cable, and it can be confusing if the cable they're holding doesn't match the description of it this page.

content/tutorials/kit-assembly.md
```diff
@@ -0,0 +1,57 @@
...
+title: Connecting your kit
+---
+
+## The bits you should have
+- Raspberry Pi (the board in a plastic case with many USB ports)
+- Power Board (the board in a plastic case with many green sockets and two buttons)
+- Motor Board (the other board in a plastic case)
+- Arduino (The blue, exposed board with a metal USB-B connector)
+- Arduino Servo Shield (another blue board, hopefully on top of the first)
+- 1 x 3.81mm Camcon (for the Arduino Servo Shield)
```
kierdavis: I think saying something like "green connector" instead of "Camcon" would be much clearer (despite being less precise and a bit of a mouthful). Not sure though - thoughts?

content/tutorials/kit-assembly.md
```diff
@@ -0,0 +1,57 @@
...
+- 2 x 5mm Camcon (for the Motors)
+- 1 x 5mm Camcon with wire loop (attached to the power board)
+- 2 x 7.5mm Camcon (for the Motor Board)
+- USB Hub
+- USB Webcam
+- 2 x Black USB-A to micro-usb cables
+- Beige USB-A to USB-B cable (for the Arduino)
+- camcon to micro-usb cable (for the Raspberry Pi)
+- Pliers
```
kierdavis: I don't think we've yet decided whether we plan to give out tools and/or motors to teams for SB2018, but I imagine it's unlikely.

I think it's valid to leave these in here for now, so long as we add a Trello card to update it once we've made a decision.

content/tutorials/kit-assembly.md
kierdavis: Alternatively, maybe it's better to just define what a "camcon" is near the top of the document (a picture would be nice too).

content/tutorials/kit-assembly.md
RealOrangeOne: Will card that up now

content/tutorials/kit-assembly.md
```diff
@@ -0,0 +1,60 @@
...
+- Motor Board (the other board in a plastic case)
+- Arduino (The blue, exposed board with a metal USB-B connector)
+- Arduino Servo Shield (another blue board, hopefully on top of the first)
+- 1 x 3.81mm Camcon (for the Arduino Servo Shield)
+- 2 x 5mm Camcon (for the Motors)
+- 1 x 5mm Camcon with wire loop (attached to the power board)
+- 2 x 7.5mm Camcon (for the Motor Board)
+- USB Hub
+- USB Webcam
+- 2 x Black USB-A to micro-usb cables
```
kierdavis: @RealOrangeOne there's a colour to remove here too (sorry, should have been clearer earlier).

content/tutorials/kit-assembly.md
kierdavis: Resolved by 65e281d503b7582c24d00ad43aaf27220a331fca.

PR #31, opened 2017-11-04
Hardware updates

RealOrangeOne:
- Improve the content in power and motor board docs.
- Structure placeholder pages for future content
- Add battery documentation

#### Todo
- [x] Charging
- [x] Pi
- [x] Sprinkle more warning messages in battery docs, can't be too careful!

PR #32, opened 2017-11-07
battery charging: specify which charging mode to use, and add note…

kierdavis:
…about an alternative mode.

Follows on from comments left on #31.


Adimote:
:shipit:

PR #35, opened 2017-11-20
Add docs for power board buzzing

RealOrangeOne:
🐝 🐝 🐝 🐝 🐝 🐝 🐝 🐝 🐝 

API reference is : https://github.com/sourcebots/robot-api/pull/30


content/api/power-board.md
```diff
@@ -19,4 +19,28 @@ r.power_board.power_off()
 `power_on` is called when you setup your robot, so this doesn't need to be called manually. The ports will come on automatically as soon as your robot is ready, before the start button is pressed.
 {{% /notice %}}
 
+## Buzzing :bee:
```
RealOrangeOne: Yes, the docs have emoji support! 🎉

content/api/power-board.md
```diff
@@ -19,4 +19,28 @@ r.power_board.power_off()
...
+
+The power board has a piezo buzzer which can beep.
+
+The `buzz` function accepts multiple parameters, depending on what you want to play. The first argument is the duration of the beep, in seconds. The later arguments are either the note you want to play, or the buzzers frequency.
+
+
+{{% notice note %}}
+You should only pass `frequency` __or__ `note`. If you pass both, `note` is used, and `frequency` is silently ignored.
```
prophile: We should do something about this!

content/api/power-board.md
```diff
@@ -19,4 +19,28 @@ r.power_board.power_off()
 `power_on` is called when you setup your robot, so this doesn't need to be called manually. The ports will come on automatically as soon as your robot is ready, before the start button is pressed.
 {{% /notice %}}
 
+## Buzzing :bee:
+
+The power board has a piezo buzzer which can beep.
+
+The `buzz` function accepts multiple parameters, depending on what you want to play. The first argument is the duration of the beep, in seconds. The later arguments are either the note you want to play, or the buzzers frequency.
```
prophile: Worth mentioning that the frequency is in Hz? (Rather than being angular frequency which would be another reasonable choice).

prophile: Maybe something about the frequency range too? It's obvious to us that the piezo can only sensibly buzz in audible frequencies but that isn't necessarily obvious to teams.

RealOrangeOne: I would really hope that's obvious, but i'll add something just in case

content/api/power-board.md
RealOrangeOne: Well, we can change this in `robot-api`, maybe throw if both values are provided might be a nice idea

RealOrangeOne: Fixed in https://github.com/sourcebots/robot-api/pull/31

content/api/power-board.md
```diff
@@ -19,4 +19,29 @@ r.power_board.power_off()
 `power_on` is called when you setup your robot, so this doesn't need to be called manually. The ports will come on automatically as soon as your robot is ready, before the start button is pressed.
 {{% /notice %}}
 
+## Buzzing :bee:
+
+The power board has a piezo buzzer which can buzz.
+
+The `buzz` function accepts multiple parameters, depending on what you want to play. The first argument is the duration of the beep, in seconds. The later arguments are either the note you want to play, or the buzzers frequency (in Hz).
```
PeterJCLaw: s/buzzers/buzzer's/, though I actually think it would be clearer phrased as "frequency of the buzz you want" or similar (something about it being a future action makes the use of "of" in the latter feel nicer to me than the possessive in the former)

content/api/power-board.md
```diff
@@ -19,4 +19,29 @@ r.power_board.power_off()
...
+
+Theoretically, the piezo buzzer will buzz at any provided frequency, however humans can only hear between [20Hz and 20000Hz](https://en.wikipedia.org/wiki/Hearing_range#Humans).  
+
+{{% notice info %}}
+To help prevent possible errors, you must specify which argument you're passing. Simply calling `buzz(2, 'c')` will raise a `TypeError`
```
PeterJCLaw: Should this end with a full stop? (I'm not sure if this is an intentional omission -- we seem to be missing full stops in a number of these `notice_info`s)

content/api/power-board.md
```diff
@@ -23,3 +23,30 @@ r.power_board.power_on()
...
+
+# Buzz for 2 seconds at 400Hz
+r.power_board.buzz(2, frequency=400)
+
+# Will fail!
+r.power_board.buzz(2, 400)
+```
+
+{{% notice warning %}}
+Passing an invalid note will raise a `KeyError`. You can see all the supported notes by looking at the power boards `BUZZ_NOTES` variable: `print(r.power_board.BUZZ_NOTES)`
```
kierdavis: Can we list these in the docs somewhere too? As it stands, users have to put this code snippet onto a USB stick, boot the robot up and examine the log in order to see what notes they can use with this function.

RealOrangeOne: I dont like documenting things like that, as they end up out of sync. Directly linking to the code isnt really ideal either

kierdavis: Directly linking to the code isn't too bad IMO, as long as the code is clear (which in this case I feel it is - it's just a list of values). It also encourages users to dig around our codebases, which is probably a good thing.

RealOrangeOne: GitHub only allows linking to line nunmbers, so again there's a place it can get out of sync

kierdavis: We're already asking them to look for the `BUZZ_NOTES` variable. If it's an obvious-looking array near the top of the file then it should be easy for people to find.

RealOrangeOne: Hmm, i guess just linking to the file might work too

PeterJCLaw: Yes, the docs can end up out of sync; that's the perennial downfall of manually written docs. Linking to code or suggesting that users print a variable for a list of valid values are unfriendly, even for experienced developers. I'm strongly in favour of listing this sort of thing in the docs.

content/api/power-board.md
```diff
@@ -23,3 +23,30 @@ r.power_board.power_on()
...
+
+## Buzzing :bee:
+The power board has a piezo buzzer which can buzz.
+
+The `buzz` function accepts multiple parameters, depending on what you want to play. The first argument is the duration of the beep, in seconds. The later arguments are either the note you want to play, or the frequency of the buzzer (in Hz). You have to specify which of note or frequency you're passing using a keyword argument, your code will fail otherwise.
+
+Theoretically, the piezo buzzer will buzz at any provided frequency, however humans can only hear between [20Hz and 20000Hz](https://en.wikipedia.org/wiki/Hearing_range#Humans).
+
+{{% notice tip %}}
+Calling `buzz` is non-blocking. Meaning it doesn't actually wait for the piezo to stop buzzing before continuing with your code. If you want duration to wait for the buzzing to stop, add a `sleep` afterwards!
```
PeterJCLaw: s/. Meaning/This means that/, please.

PeterJCLaw: We should clarify what a `sleep` is -- unless you happen to know about `time.sleep()` then just `sleep` doesn't mean much.

PR #36, opened 2017-11-25
Add license file respecting history from imported content

RealOrangeOne:
This is MIT with additional contributors, due to importing kit and
board docs and whirlwind tour from https://github.com/srobo/docs.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,29 @@
+Copyright (c) 2010-2017 Chris Kirkham <chrisjameskirkham@gmail.com>
+	      		Richard Barlow <richard@richardbarlow.co.uk>
+			Andrew Busse <andybusse@studentrobotics.org>
+			Robert Spanton <rspanton@zepler.net>
+			Peter Law <PeterJCLaw@gmail.com>
+			Alastair Lynn <arplynn@gmail.com>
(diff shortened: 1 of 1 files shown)


LICENSE
```diff
@@ -0,0 +1,29 @@
+Copyright (c) 2010-2017 Chris Kirkham <chrisjameskirkham@gmail.com>
+	      		Richard Barlow <richard@richardbarlow.co.uk>
```
PeterJCLaw: It'd be good if we could have consistent whitespace here (I suggest using spaces since we want to align with the top line).

PeterJCLaw:
@RealOrangeOne I've been thinking about this some more and I'm increasingly of the opinion that we should be clear about the license of this work as separate from the bits pulled in from elsewhere. I think we maybe want to go with something that clarifies the different contributions.

I've also run a blame on srweb's docs, which yields quite a few more contributors (I'll be upstreaming the result shortly).

PeterJCLaw:
@RealOrangeOne why was this frozen (and what does that mean)?

RealOrangeOne:
it was frozen as a note for me not to merge at kickstart. License things came up in the last meeting, and are being researched into, apparently

PR #39, opened 2017-12-03
FAQs

RealOrangeOne:
If a team asks a question, we should probably word it into the docs somehow, or make it more obvious.

If a question doesn't fit well into a section, add the question to this page


config.yml
```diff
@@ -11,7 +11,7 @@ params:
     disableLanguageSwitchingButton: true
     disableAssetsBusting: true
     editURL: "https://github.com/sourcebots/docs/tree/master/content/"
-    ordersectionsby: "title"
+    ordersectionsby: "weight"
```
sedders123: Does this fall back to 'title' for all sections with the same weight?

content/faqs/_index.md
```diff
@@ -0,0 +1,18 @@
+---
+title: FAQs
+weight: 10
+---
+
+Teams often ask us questions, so we've put the most common ones below, so others can learn from them.
+
+{{% expand "Do these FAQs work?" %}}
```
sedders123: Is it worth coming up with a couple of 'real' FAQs instead of just dummy data?

config.yml
RealOrangeOne: Yes. It was just a hack to move the FAQs section to the bottom of the list

content/faqs/_index.md
kierdavis: I agree, right now this page looks like a test page (even though we intend to add useful content at a later date)

RealOrangeOne: Well, that's technically what it is. But i'll try and come up with some generic things to fill it out a bit

PR #40, opened 2017-12-03
Warn when changing motor power too quickly

RealOrangeOne:
We noticed the current limit was hit whenever motors were changed too suddenly. Whilst this is known by us, it wasn't written in the docs


content/api/motor-board.md
```diff
@@ -75,3 +75,8 @@ from robot import COAST
 
 r.motor_board.m1.voltage = COAST
 ```
+
+{{% notice warning %}}
+Changing motor speed by more than 1 in a single operation (e.g. `-1` to `0`, `1` to `-1` etc) will likely trigger the current protection and your robot will shut down with a distinct beeping noise.
```
sedders123: Could possibly use 'the' between the first two words

PR #42, opened 2017-12-03
Add list of additionally installed libraries

RealOrangeOne:
Add list of packages outside the standard library available if required.

Actually implemented in https://github.com/sourcebots/pi-image/pull/7


RealOrangeOne:
Once the related implementation PR is merged, adapt to read libraries directly from the file in the `pi-image` repo

content/api/included-libraries.md
```diff
@@ -0,0 +1,8 @@
+---
+title: Included Libraries
+---
+
+Python already comes with plenty of [built-in libraries](https://docs.python.org/3.4/py-modindex.html) to use, However there are some extra ones installed which may be of us:
```
sedders123: Once #45 is merged could `https://docs.python.org/3.4/` be replaced with the python_version variable?

sedders123: 'may be of us' should be 'may be of use' ? also 'However' should not be capitalised after a comma

.spelling
```diff
@@ -26,3 +26,5 @@ square-ish
 83x99x24mm
 4mm
 2.2A.
+numpy
+scipy
```
sedders123: trailing newline?

PR #45, opened 2017-12-07
Specify version of python

RealOrangeOne:
Specify the version of python we're using. The default in stretch is `3.5`, so that's what we're using!

### DIFF
--- a/content/tutorials/python-whirlwind-tour.md
+++ b/content/tutorials/python-whirlwind-tour.md
@@ -1,7 +1,7 @@
 ---
 title: "Python: A Whirlwind Tour"
 ---
-In this tutorial, we'll introduce the basic concepts of programming, which will be central to the programs that you will run on your robot. There are many different languages in which computers can be programmed, all with their advantages and disadvantages, but we use [Python](https://www.python.org/), specifically version 3. We chose Python because it's easy to learn, but also elegant and powerful.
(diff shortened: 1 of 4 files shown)


content/tutorials/python-whirlwind-tour.md
```diff
@@ -1,7 +1,7 @@
 ---
 title: "Python: A Whirlwind Tour"
 ---
-In this tutorial, we'll introduce the basic concepts of programming, which will be central to the programs that you will run on your robot. There are many different languages in which computers can be programmed, all with their advantages and disadvantages, but we use [Python](https://www.python.org/), specifically version 3. We chose Python because it's easy to learn, but also elegant and powerful.
+In this tutorial, we'll introduce the basic concepts of programming, which will be central to the programs that you will run on your robot. There are many different languages in which computers can be programmed, all with their advantages and disadvantages, but we use [Python](https://www.python.org/). We chose Python because it's easy to learn, but also elegant and powerful.
```
PeterJCLaw: Not sure why you've removed  the detail here; having it duplicated feels quite reasonable to me.

If we're worried about things getting out of step we could:
 * Refer to it as "Python 3" in most places and only specify the actual version number once
 * Define a constant for the version number and then reference that everywhere

RealOrangeOne: I like the idea of a constant, will add

layouts/shortcodes/python_version.html
```diff
@@ -0,0 +1 @@
+{{ .Site.Params.pythonVersion }}
```
PeterJCLaw: I'm not sure I see the point of this indirection?

RealOrangeOne: It's just a little syntactic sugar. Means you can use `{{% python_version %}}` where it's needed rather than `{{ .Site.Params.pythonVersion }}

PR #46, opened 2017-12-09
Add table of kit we're shipping

RealOrangeOne:
- Add pi to list
- Ship spare camcons
- We're now shipping 2 batteries


kierdavis:
Surely the Pi should be on this list?

I'd also recommend adding the Pi power cable explicitly (even though it's mentioned in the comments for another item).

RealOrangeOne:
You are correct on both, not sure how I missed those

kierdavis:
It's also just occurred to me that we should probably ship them spare camcons. They're comparatively easy to lose and it seems silly for them to get blocked while we ship them a replacement.

PR #49, opened 2017-12-11
Vision page

antonnikitin97:


content/api/vision.md
```diff
@@ -0,0 +1,12 @@
+---
+title: Vision
+---
+
+The main way your robot will sense the outside world is through the webcam which is used to read the AprilTag markers.
+
+{{% notice info %}}
```
RealOrangeOne: This is already defined at the intro to the API section, it's not needed here

antonnikitin97: Fair enough, but I feel the page needs an introductory sentence :) I'll remove the part about the AprilTags..

content/api/vision.md
```diff
@@ -0,0 +1,61 @@
...
+```python
+camera = r.camera
+```
+The software will process the image and return any markers it sees into an array sorted by distance.
+
+```python
+markers = r.camera.see()
+```
+
+You can then iterate over the `markers` array just as you would any other list:
```
RealOrangeOne: Using both `array` and `list` feels wrong here. Maybe drop `array`, it makes sense as just _iterate over the markers just as you would any other list_

content/api/vision.md
```diff
@@ -0,0 +1,61 @@
...
+for marker in markers:
+  # Do robot stuff
+```
+
+## Marker properties
+The markers that are contained within the array have some useful properties:
+
+  - `is_wall_marker()` - returns whether or not the marker is a wall marker.
+  - `is_token_marker()` - returns whether or not the marker is a token marker.
+  - `id()` - returns the id of the marker
```
RealOrangeOne: This is a property, it's not called as a function. As are `pixel_center`, `polar` and `cartesian`, which is also spelt wrong

content/api/vision.md
```diff
@@ -0,0 +1,61 @@
...
+  - `pixel_centre()` - returns a [tuple](https://www.tutorialspoint.com/python/python_tuples.htm) containing the pixel coordinates of the centre of the marker.
+  - `polar()` - returns a PolarCoord object containing the position of the marker in the Polar coordinate system.
```
RealOrangeOne: Wrap the object type  in backticks (`), looks far nicer!

content/api/vision.md
```diff
@@ -0,0 +1,61 @@
...
+## Cartesian and Polar Coordinate Systems
+The markers' position can be represented using both the Polar and Cartesian coordiate system. The two objects `CartCoord` and `PolarCoord` hold the position information for both of these systems.
+
+{{% notice info %}}
+[Cartesian coordinates](https://en.wikipedia.org/wiki/Cartesian_coordinate_system) are ones that are defined using a unique set of values (typically x, y and z) to specify a point in space. [Polar coordinates](https://en.wikipedia.org/wiki/Polar_coordinate_system) are represented by a distance from a set point and an angle from a set direction.
+{{% /notice %}}
+
+The `CartCoord` object has the following properties:
+
+  - `x()` - returns the x coordinate of the marker in Cartesian space, x is defined as being positive in the right direction.
```
RealOrangeOne: these are properties, not called as a function

content/api/vision.md
```diff
@@ -0,0 +1,61 @@
...
+  - `y()` - returns the y coordinate of the marker in Cartesian space, y is defined as being positive in the up direction.
+  - `z()` - returns the z coordinate of the marker in Cartesian space, it is defined as the distance from the camera to the marker in **meters**
```
RealOrangeOne: Maybe remove the unit of measurement from just the `z`, and add it as a notice

content/api/vision.md
```diff
@@ -0,0 +1,61 @@
...
+  
+{{% notice info %}}
```
RealOrangeOne: This notice is unnecessary, as the tuple youre quoting is never actually used. You can remove this notice entirely

content/api/vision.md
```diff
@@ -0,0 +1,61 @@
...
+  - `cartersian()` - returns a CartCoord object containing the position of the marker in the Cartesian coordinate system.
+
+## Cartesian and Polar Coordinate Systems
+The markers' position can be represented using both the Polar and Cartesian coordiate system. The two objects `CartCoord` and `PolarCoord` hold the position information for both of these systems.
+
+{{% notice info %}}
+[Cartesian coordinates](https://en.wikipedia.org/wiki/Cartesian_coordinate_system) are ones that are defined using a unique set of values (typically x, y and z) to specify a point in space. [Polar coordinates](https://en.wikipedia.org/wiki/Polar_coordinate_system) are represented by a distance from a set point and an angle from a set direction.
+{{% /notice %}}
+
+The `CartCoord` object has the following properties:
```
RealOrangeOne: You can probably generalize this entire section. It has 3 properties (`x`, `y` and `z`), measured in meters, and the axis are relative to your robot (or its webcam anyway)

content/api/vision.md
```diff
@@ -0,0 +1,61 @@
...
+  - `y()` - returns the y coordinate of the marker in Cartesian space, y is defined as being positive in the up direction.
+  - `z()` - returns the z coordinate of the marker in Cartesian space, it is defined as the distance from the camera to the marker in **meters**
+  
+{{% notice info %}}
+(0,0,1) would be exactly 1 metre in front of the marker, with the camera looking at the centre.
+{{% /notice %}}
+  
+The `PolarCoord` object has the following properties:
+ - `distance_meters()` - returns the distance from the camera to the marker in **meters**.
+ - `rot_x_rad()` and `rot_x_deg` - returns the angle of rotation in radians/degrees about the x axis. The x axis runs from left to right in front of the camera. A positive angle would be one where the marker is leaning toward the camera.
```
RealOrangeOne: Stating which direction the axis runs in is probably unnecessary.

content/api/vision.md
```diff
@@ -0,0 +1,61 @@
...
+  - `x()` - returns the x coordinate of the marker in Cartesian space, x is defined as being positive in the right direction.
+  - `y()` - returns the y coordinate of the marker in Cartesian space, y is defined as being positive in the up direction.
+  - `z()` - returns the z coordinate of the marker in Cartesian space, it is defined as the distance from the camera to the marker in **meters**
+  
+{{% notice info %}}
+(0,0,1) would be exactly 1 metre in front of the marker, with the camera looking at the centre.
+{{% /notice %}}
+  
+The `PolarCoord` object has the following properties:
+ - `distance_meters()` - returns the distance from the camera to the marker in **meters**.
```
RealOrangeOne: This is a property, it's not called as a function. As are all the keys on this object

content/api/vision.md
kierdavis: Would be good to explain what a "marker ID" is anyway

content/api/vision.md
antonnikitin97: as in remove 'pixel_centre' or remove the tuple link?

RealOrangeOne: Remove the _(0,0,1) would be exactly 1 metre in front of the marker, with the camera looking at the centre._

content/api/vision.md
```diff
@@ -31,31 +31,35 @@ for marker in markers:
...
+All distances in this section are measured in meters.
+{{% /notice %}}
+
 The `CartCoord` object has the following properties:
 
-  - `x()` - returns the x coordinate of the marker in Cartesian space, x is defined as being positive in the right direction.
-  - `y()` - returns the y coordinate of the marker in Cartesian space, y is defined as being positive in the up direction.
-  - `z()` - returns the z coordinate of the marker in Cartesian space, it is defined as the distance from the camera to the marker in **meters**
-  
+  - `x` - returns the x coordinate of the marker in Cartesian space, x is defined as being positive in the right direction.
```
RealOrangeOne: You can probably generalize this entire section. It has 3 properties (x, y and z), measured in meters, and the axis are relative to your robot (or its webcam anyway)

content/api/vision.md
```diff
@@ -0,0 +1,61 @@
...
+
+## Using the camera to search for markers
+If you have a webcam connected, you can use `r.camera.see()` to take a picture. 
+
+If you consider it necessary you can assign the camera object to a variable just like `Robot`:
+
+```python
+camera = r.camera
+```
+The software will process the image and return any markers it sees into an array sorted by distance.
```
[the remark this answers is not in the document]
antonnikitin97: I'm not sure XD I will ask :)

content/api/vision.md
```diff
@@ -0,0 +1,72 @@
+---
+title: Vision
+---
+
+Your robot uses its webcam to see markers.
+
+{{% notice info %}}
```
RealOrangeOne: This should be removed

content/api/vision.md
```diff
@@ -0,0 +1,72 @@
...
+  
+The `PolarCoord` object has the following properties:
+ - `distance_meters` - returns the distance from the camera to the marker.
+ - `rot_x_rad` and `rot_x_deg` - returns the angle of rotation in radians/degrees about the x axis. The x axis runs horizontally in front of the camera. A positive angle would be one where the marker is leaning toward the camera.
+ - `rot_y_rad` and `rot_x_deg` - returns the angle of rotation in radians/degrees about the y axis. The y axis runs vertically in front of the camera. A positive angle would be one where the marker is rotated clockwise about the y axis looking from the top of the marker.
+
+## Marker IDs
+Each marker used in the game has a unique ID that corresponds to the marker type. You can use this within your code to determine what sort of marker you're looking at and make decisions based off of it.
+
+There are two main marker types in the game WALL and TOKEN:
```
RealOrangeOne: You should state where these constants can be imported from, and wrap it in backticks to make it look nicer

content/api/vision.md
```diff
@@ -0,0 +1,72 @@
...
+  - `y` - returns the y coordinate of the marker in Cartesian space, y is defined as being positive in the up direction.
+  - `z` - returns the z coordinate of the marker in Cartesian space, it is defined as the distance from the camera to the marker.
+
+{{% notice info %}}
+(0,0,1) would be exactly 1 metre in front of the marker, with the camera looking at the centre.
+{{% /notice %}}
+  
+The `PolarCoord` object has the following properties:
+ - `distance_meters` - returns the distance from the camera to the marker.
+ - `rot_x_rad` and `rot_x_deg` - returns the angle of rotation in radians/degrees about the x axis. The x axis runs horizontally in front of the camera. A positive angle would be one where the marker is leaning toward the camera.
```
RealOrangeOne: Dont state where the axis runs, it's unnecessary

content/api/vision.md
```diff
@@ -0,0 +1,72 @@
...
+All distances in this section are measured in meters.
+{{% /notice %}}
+
+The `CartCoord` object has the following properties:
+
+  - `x` - returns the x coordinate of the marker in Cartesian space, x is defined as being positive in the right direction.
+  - `y` - returns the y coordinate of the marker in Cartesian space, y is defined as being positive in the up direction.
+  - `z` - returns the z coordinate of the marker in Cartesian space, it is defined as the distance from the camera to the marker.
+
+{{% notice info %}}
```
RealOrangeOne: Remember to remove this

content/api/vision.md
antonnikitin97: I disagree, I feel that the more explicit we are the more helpful the docs will be.

RealOrangeOne: We dont need to define what an X-axis is though

content/api/vision.md
```diff
@@ -0,0 +1,70 @@
+---
+title: Vision
+---
+
+{{% notice info %}}
```
RealOrangeOne: I think you removed the wrong thing. I meant remove this notice, not the opening sentence 😂

content/api/vision.md
```diff
@@ -0,0 +1,70 @@
...
+ - `distance_meters` - returns the distance from the camera to the marker.
+ - `rot_x_rad` and `rot_x_deg` - returns the angle of rotation in radians/degrees about the x axis. The x axis runs horizontally in front of the camera. A positive angle would be one where the marker is leaning toward the camera.
+ - `rot_y_rad` and `rot_x_deg` - returns the angle of rotation in radians/degrees about the y axis. The y axis runs vertically in front of the camera. A positive angle would be one where the marker is rotated clockwise about the y axis looking from the top of the marker.
+
+## Marker IDs
+Each marker used in the game has a unique ID that corresponds to the marker type. You can use this within your code to determine what sort of marker you're looking at and make decisions based off of it.
+
+There are two main marker types in the game `WALL` and `TOKEN`:
+  - `WALL` markers are ones that are on the walls of the arena and have marker IDs from 0 - 27
+  - `TOKEN` markers are ones that are the on the boxes and have marker IDs from 44 - 83
```
RealOrangeOne: This now needs changing, as of https://github.com/sourcebots/robot-api/pull/37/. I'd also suggest referencing these constants

content/api/vision.md
antonnikitin97: Well heck!

content/api/vision.md
antonnikitin97: This is true, but from the perspective of people that may not understand the concepts very clearly, I think it helps to be as explicit as possible. After all, docs are supposed to be in depth :)

content/api/vision.md
antonnikitin97: I have now implemented the changes :) 👍

content/api/vision.md
kierdavis: I agree with Anton here. There's not really "one unified standard" of three dimensional coordinate system, so it's useful if not vital to say which directions the axes of our coordinate system point.

content/api/vision.md
```diff
@@ -0,0 +1,76 @@
+---
+title: Vision
+---
+
+Your webcam is how your robot will see markers.
+
+{{% notice info %}}
+Throughout this page it is assumed you have a `Robot` object defined as `r`.
```
RealOrangeOne: Remove this notice

antonnikitin97: I have implemented this change :)

content/api/vision.md
```diff
@@ -0,0 +1,72 @@
...
+
+{{% notice info %}}
+All distances in this section are measured in meters.
+{{% /notice %}}
+
+The `CartCoord` object has the following properties, where each coordinate is in Cartesian space:
+
+  - `x` - returns the x coordinate, defined as being positive in the right direction.
+  - `y` - returns the y coordinate, defined as being positive in the up direction.
+  - `z` - returns the z coordinate, defined as the distance from the camera to the marker.
```
PeterJCLaw: I don't think this is correct as it implies that the axes are unique for each marker. I think you mean "the distance forwards in front of the camera". 

The current description implies that this value is the absolute distance to the marker in a direct line, rather than the distance forwards. For a marker directly in front of the robot this makes little difference, but could have quite a big impact for a marker off to one side.

content/api/vision.md
```diff
@@ -0,0 +1,72 @@
...
+
+{{% notice info %}}
+For example, if x, y and z were 0, 0 and 1 respectively. This would correspond to a marker being exactly 1 metre in front of the camera, looking at the centre.
```
PeterJCLaw: "looking at the centre" sounds weird here.

content/api/vision.md
PeterJCLaw: We've defined our axes above when describing the location of the marker. We shouldn't need to repeat that.

content/api/vision.md
```diff
@@ -0,0 +1,72 @@
...
+{{% /notice %}}
+  
+The `PolarCoord` object has the following properties:
+ - `distance_meters` - returns the distance from the camera to the marker.
+ - `rot_x_rad` and `rot_x_deg` - returns the angle of rotation in radians/degrees about the x axis. The x axis runs horizontally in front of the camera. A positive angle would be one where the marker is leaning toward the camera.
+ - `rot_y_rad` and `rot_x_deg` - returns the angle of rotation in radians/degrees about the y axis. The y axis runs vertically in front of the camera. A positive angle would be one where the marker is rotated clockwise about the y axis looking from the top of the marker.
```
PeterJCLaw: typo (I hope): s/`rot_x_deg`/`rot_y_deg`/ here.

content/api/vision.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Marker IDs
+Each marker used in the game has a unique ID that corresponds to the marker type. You can use this within your code to determine what sort of marker you're looking at and make decisions based off of it.
+
+There are three main marker types in the game `WALL`, `TOKEN` and `TOKEN_ZONE_x` where `x` is the starting zone in the arena:
+  - `WALL`- Markers that are on the walls with IDs from 0 - 27
```
PeterJCLaw: Why is `WALL` here in backticks? Are we describing a constant which can be used in code? If so, we should be explicit about that. Otherwise we should remove the implication that it's code.

I also dislike that we're specifying the ranges at all -- it couples this page to the game much more than is needed, but misses out some game specifics. I'd much rather drop this and instead encourage use of either named constants in the code (for the type of the marker), or the methods like `is_wall_marker()` on the `Marker` instance.

content/api/vision.md
RealOrangeOne: This was fixed in a later PR: https://github.com/sourcebots/docs/pull/62

content/api/vision.md
RealOrangeOne: I'm not sure it's possible to fully decouple the pages from the game itself isn't especially possible. I think the best solution here however will be to create a sub-page off vision for the game-specific components of Vision

RealOrangeOne:
@PeterJCLaw very good catches. I'll go through and make these changes / update the guide now!

content/api/vision.md
PeterJCLaw: Sure. I've realised that it's not so much the game specifics which I'm not enjoying here, as the inconsistency between whether we include game information and the level of detail we give.

I think it's ok for this page (without the need for a sub-page) to be coupled to the game, though I'd rather not expose the actual ids used since they should be an implementation detail on our side. I'd even go so far as to suggest that the rules also shouldn't include the ids (and instead maybe reference a constant).

I don't think we should remove the ids from the rules now that they're published, though I don't see any benefit to providing them in the docs.

I suggest that we:
 - remove the ids from here
 - clarify whether or not these are constants
 - add some documentation about the markers which are on the columns

kierdavis:
@PeterJCLaw apologies - I wanted to get this shipped at kickstart so may have missed some typos and grammar errors when reviewing. Looks like Jake has addressed your comments in #67.

PeterJCLaw:
@kierdavis no worries; I agree that having something live for Kickstart (rather than waiting for more review) was the right call here.
I've not yet worked out how in github to "review" code/etc. which is already merged; hence reviewing on the PR which introduced it.
I'll bear in mind that I should clarify that intent in future.

PR #50, opened 2017-12-12
Warn turning off power outputs prevents motor board use

RealOrangeOne:
Can't use motors without power!

### DIFF
--- a/content/api/power-board.md
+++ b/content/api/power-board.md
@@ -11,12 +11,15 @@ main_board = r.power_board
 ## Power Outputs
 The power boards 6 outputs can be turned on and off using the `power_on` and `power_off` functions respectively.
 
-```python
-r.power_board.power_off()
-```
-
 {{% notice note %}}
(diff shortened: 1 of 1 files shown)


content/api/power-board.md
```diff
@@ -11,12 +11,15 @@ main_board = r.power_board
...
 `power_on` is called when you setup your robot, so this doesn't need to be called manually. The ports will come on automatically as soon as your robot is ready, before the start button is pressed.
 {{% /notice %}}
 
+```python
+r.power_board.power_off()
+r.power_board.power_on()
+```
 
+{{% notice warning %}}
+Because the power board is powered through these power outputs, whilst the power is off, you won't be able to control your motors. This will register as a missing board and your code will break if you try and control them.
```
sedders123: Is this line meant to read, 'Because the **motor** board'?

RealOrangeOne: It is 🤦‍♂️

PR #57, opened 2017-12-16
added explanitory text to index page

antonnikitin97:
felt it would be nice to fill the blank page with something


RealOrangeOne:
Addressed in #76

PR #59, opened 2017-12-16
Document usage of digital pins 0 and 1

ninjafrostpn:
Or rather not. Students will encounter weird behaviour if they try to use them, as the Arduino uses these for serial comms.
Also general capitalisation and things.

### DIFF
--- a/content/kit/servo-assembly/GPIO.md
+++ b/content/kit/servo-assembly/GPIO.md
@@ -2,7 +2,7 @@
 title: GPIO
 ---
 
-The arduino allows you to connect your kit to your own electronics. It has 14 digital I/O pins, and 6 analogue. The analogue pins can read an analogue signal from 0 to 5V. The board also has a couple of ground pins, as well as some pins fixed at 3.3V and 5V output.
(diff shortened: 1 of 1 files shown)


content/kit/servo-assembly/GPIO.md
```diff
@@ -2,13 +2,13 @@
...
+The Arduino allows you to connect your kit to your own electronics. It has 14 digital I/O pins, and 6 analogue. The analogue pins can read an analogue signal from 0 to 5V. The board also has a couple of ground pins, as well as some pins fixed at 3.3V and 5V output.
 
 The pin layout of the servo assembly is the same as an Arduino uno.
 
 ![Pin Map](https://raw.githubusercontent.com/Bouni/Arduino-Pinout/master/Arduino%20Uno%20R3%20Pinout.png?width=50pc)
 
 
 {{% notice note %}}
-The servo hat communicates using 2 of the analogue input pins (4 and 5), and so these are reserved. Using these pins will cause the servo hat to behave weirdly.
+The servo hat communicates with the Arduino using two of the analogue input pins (4 and 5) and it communicates with the Pi using 2 of the digital IO pins (0 and 1), so these four pins are reserved. Using these pins will cause the servo hat to behave unusually.
```
kierdavis: Technically, it's the Arduino that communicates with the Pi using pins 0 and 1, while this sentence implies its the servo hat that communicates with the Pi. Although this doesn't make a different to the user it'd be good to be more precise here.

kierdavis: Fixed this in #70

PR #66, opened 2017-12-17
tutorials/basic-movement: divide by a hundred

kierdavis:
Some of the code samples still used old the -100..100 range for motor power.
This commit updates them to use the new -1..1 range.

This was spotted by a team at kickstart.

### DIFF
--- a/content/tutorials/basic-movement.md
+++ b/content/tutorials/basic-movement.md
@@ -130,28 +130,28 @@ while True:
 
     # ^^ code from before ^^
 
-    # power up to 70 (from 10)
+    # power up to 0.7 (from 0.1)
     for pwr in range(10, 80, 10):
-        r.motor_board.m0 = pwr
+        r.motor_board.m0 = pwr / 100.0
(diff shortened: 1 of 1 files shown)


content/tutorials/basic-movement.md
```diff
@@ -130,27 +130,27 @@ while True:
 
     # ^^ code from before ^^
 
-    # power up to 70 (from 10)
-    for pwr in range(10, 80, 10):
+    # power up to 0.7 (from 0.1)
+    for pwr in range(0.1, 0.8, 0.1):
```
RealOrangeOne: Unfortunately, it's not this simple, `range` only accepts integers. You'll probably need to keep the range and do a divide in code.

kierdavis: Good spot - didn't know that was they case, but [that is indeed what the docs say](https://docs.python.org/3/library/stdtypes.html#typesseq-range).

kierdavis:
cheers :+1:

PR #67, opened 2017-12-17
Vision cleanup

RealOrangeOne:
Addresses comments in https://github.com/sourcebots/docs/pull/49#pullrequestreview-84009406

Also does some general cleanup / code correction

(Although the diff might look that way, it was _far_ from a complete rewrite!)


content/api/vision/_index.md
```diff
@@ -0,0 +1,26 @@
...
+markers = r.camera.see()
+```
+
+## Marker properties
+The markers in the list have some useful properties:
+
+- `is_wall_marker()` - returns whether or not the marker is a [wall marker](marker-ids/#wall-markers).
+- `is_token_marker()` - returns whether or not the marker is a [token marker](marker-ids/#token-markers).
+- `id` - returns the [id](marker-ids) of the marker.
+- `pixel_centre` - returns the pixel coordinates of the centre of the marker.
```
kierdavis: I think it'd be a good idea to define "pixel coordinates", as the meaning of the phrase by itself isn't very obvious.

content/api/vision/_index.md
```diff
@@ -0,0 +1,26 @@
...
+- `polar` - returns details of the position of the marker in the [Polar](coordinates/#polar-coordinates) coordinate system.
+- `cartesian` - returns details of the position of the marker in the [Cartesian](coordinates/#cartesian-coordinates) coordinate system.
+
+{{% notice tip %}}
+The markers' position can be represented using both the [Polar](coordinates/#polar-coordinates) and [Cartesian](coordinates/#cartesian-coordinates) coordinate system.
```
kierdavis: The grammar doesn't quite agree at the start of this sentence. Please change to either "marker's position" (the position of a single marker) or "markers' positions" (the positions of many markers).

Also, "system" -> "systems", since we're talking about more than one coordinate system.

content/api/vision/coordinates.md
```diff
@@ -0,0 +1,34 @@
...
+
+{{% notice note %}}
+All distances are measured in metres.
+{{% /notice %}}
+
+## Polar Coordinates
+The polar coordinates are taken from the position of the webcam.
+
+ - `distance_metres` - the distance from the camera to the marker.
+ - `rot_x_rad` / `rot_x_deg` - the angle of rotation in radians/degrees about the x axis. The x axis runs horizontally in front of the camera. A positive angle would be one where the marker is in front of the camera.
```
kierdavis: This second sentence seems odd - surely all markers seen by the camera must be in front of it? Perhaps a sentence like the corresponding one for the y axis would make more sense here.

content/api/vision/marker-ids.md
```diff
@@ -0,0 +1,80 @@
...
+title: Marker IDs
+---
+
+Each marker used in the game has an ID that corresponds to the marker type. You can use this within your code to determine what sort of marker you're looking at and make decisions based off of it.
+
+There are some useful constants which can be imported to do this. For example:
+
+```python
+from robot import WALL, TOKEN_ZONE_1, COLUMN_N_FACING_N
+```
```
kierdavis: Since there seems to be a lot of constants like these, I think here would be a good place to mention `from robot import *`.

content/api/vision/marker-ids.md
```diff
@@ -0,0 +1,80 @@
...
+{{% /notice %}}
+
+```python
+marker.id in COLUMN  # Is the marker a column marker?
+ 
+marker.id in COLUMN_N  # Is the marker on the North column? 
+
+marker.id in COLUMN_FACING_E  # Is the marker facing East on its column?
+ 
+marker.id in COLUMN_S_FACING_W  # Is the marker on the South column facing West?
```
kierdavis: Looking at the [source](https://github.com/sourcebots/robot-api/blob/master/robot/game_specific.py#L27), it appears `COLUMN_S_FACING_W` and the like are single marker IDs rather than sets, and so we need to use `==` rather than `in` here. @Adimote could you confirm this is the case?

Personally I'm not a fan of the inconsistency and would prefer to make `COLUMN_S_FACING_W` a set in `robot-api` instead, but I'm aware we've already shipped it.

content/api/vision/marker-ids.md
```diff
@@ -0,0 +1,80 @@
...
+```python
+marker = r.camera.see()[0]  # Only the closest marker
+
+marker.id in WALL
+>>> True 
+```
+
+To see the full range of constants, and the exact IDs they match up to, have a look at [the source](https://github.com/sourcebots/robot-api/blob/master/robot/game_specific.py).
+
+### _Wall_ Markers
```
kierdavis: What's the purpose of the emphasis on "wall" here (and similarly for the following two headings)?

content/api/vision/marker-ids.md
```diff
@@ -0,0 +1,80 @@
...
+marker.id in COLUMN_S_FACING_W  # Is the marker on the South column facing West?
+```
+{{% notice note %}}
+In the above examples, the points of the compass may be replaced with any others.
+{{% /notice %}}
+
+### _Token_ Markers
+Each token has a marker on its face. These markers can be used to work out what teams token it is.
+
+There's a constant for each zone, and an additional to work out if it's a marker token you're looking at.
```
kierdavis: I think this could be phrased a little better - how about:

> Each scoring zone has a corresponding constant that matches tokens belonging to that zone. There's also another constant that matches tokens belonging to any zone.

content/api/vision/marker-ids.md
```diff
@@ -0,0 +1,80 @@
...
+
+marker.id in TOKEN_ZONE_0  # Does this token belong to the team in zone 0?
+```
+
+{{% notice info %}}
+Each face of the token has the same marker on. However each token has different markers.
+{{% /notice %}}
+
+{{% notice tip %}}
+It's recommended to use `marker.is_token_marker()` instead of `marker.id in TOKEN`
```
kierdavis: Missing period.

content/api/vision/marker-ids.md
RealOrangeOne: I really dislike this pattern, as no only is it not obvious if / where something it defined, it also makes linting impossible

content/api/vision/marker-ids.md
RealOrangeOne: Looking at the source, youre right. I'll change to `==`. I think this is the right decision, as using `in` would imply there's more than 1 id for `COLUMN_S_FACING_W`

content/api/vision/_index.md
```diff
@@ -0,0 +1,26 @@
+---
+title: Vision
+---
+
+Your webcam is how your robot will see markers.
+
+## Searching for markers
+Assuming you have a webcam connected, you can use `r.camera.see()` to take a picture. The software will process the picture and returns a list of the markers it sees, sorted by distance, where the first element of the array is the closest marker to the camera.
```
PeterJCLaw: I realise this isn't quite a docs thing, but _why_ are we sorting the markers by distance? This seems completely gratuitous and not good API design.

antonnikitin97: Saves the students having to sort them? :) (Sorry, I don't really understand why it's bad API design)

PeterJCLaw: There are a number of reasons I don't like the pre-sorting:
1. Sorting a list can be an expensive operation -- it's always an operation which scales at least with the size of the list (`O(N)`) and often _much_ worse. By doing it inside the API we're forcing all clients to bear that cost, rather than letting them decide whether they need to pay it. Assuming that you know what your consumers want is a great way to build an API which works a little better for the case you're building for and really badly for everything else. I'd rather build an API which is generally useful and works well overall.
2. Sorting the list is _really easy_ to do if the client wants it. Our implementation is (as far as I know) purely for apparent convenience and doesn't offer any improvement over the client doing it later. If we think that the students won't be able to work out how to do it and/or want to make it easier to get to that outcome, then I suggest we include a suitable example in the docs. The precedent for this are the extensive example, how-tos and cookbooks in the main Python docs.
   * Note: if our sort was somehow _much_ more efficient then I'd be more open to the pre-sorting, though would instead encourage a flag or separate API to access it. As it is, we're literally calling the same code that the students could (see the call to `sorted` at https://github.com/sourcebots/robot-api/blob/master/robot/camera.py#L32)
3. I don't think it's a good idea for us to make our APIs too "nice" in ways that real APIs aren't as it won't teach the general skills we're actually trying to teach. It will also make it harder when the competitors encounter a less "nice" API. Of course this _doesn't_ mean that we shouldn't make the API easy to use!

content/api/vision/marker-ids.md
kierdavis: Sounds good to me. In that case I think it's be a good idea to explain *why* we use `in` for some of the constants and `==` for others (i.e. some are sets of marker IDs, others are just a single marker ID).

content/api/vision/marker-ids.md
kierdavis: Fair enough, that's fine.

content/api/vision/_index.md
kierdavis: I think a good compromise would be to add an optional argument to disable sorting. This keeps the API convenient for entry-level users.

content/api/vision/coordinates.md
RealOrangeOne: I'm not entirely sure how we'd explain this. The chances of them actually needing the `x` variant are fairly low besides getting on-ground distance. I think it's likely both need a rewrite, however I don't know what the correct wording is.

content/api/vision/_index.md
RealOrangeOne: I think the argument on computation time is fairly mute. The markers are sorted off a constant value, and with only a handful of markers, the computation and sort time for this is minimal, especially when compared to the computation time of the marker scan. 
I don't see much use in having a flag, as the chances of anyone actually _wanting_ unsorted markers is fairly remove. I suggest we either leave as-is, or have them arbitrarily sorted (however they come out `sb-vision`.

content/api/vision/coordinates.md
PeterJCLaw: I think the issue comes from us trying to define only part of the axis -- the location of the positive direction, but not the location of zero degrees.

How about saying "Angles are measured from directly above the robot around in front of the robot. Any markers the camera can see will be in the range 0° to 180°, with 90° meaning the marker is directly in front of the camera."

I'm not sure about the "directly in front of the camera" bit as it implies that markers off to the side would have a different value. However, I also don't want to say "at the same level as the camera" since that would be wrong if the camera is pointed up or down at all.

Thoughts?

content/api/vision/coordinates.md
```diff
@@ -0,0 +1,34 @@
...
+ - `rot_y_rad` / `rot_y_deg` - the angle of rotation in radians/degrees about the y axis. The y axis runs vertically in front of the camera. A positive angle would be one where the marker is rotated clockwise about the y axis looking from the top of the marker.
```
PeterJCLaw: I think we should s/in front of/through/ in the phrase "The y axis runs vertically in front of the camera" since that would be more correct.

I also think we can clarify the directions by saying something like: "markers to the right of the camera will have a positive value, markers to the left will have a negative value".

content/api/vision/coordinates.md
```diff
@@ -0,0 +1,34 @@
...
+ 
+{{% notice info %}}
+For example, if `distance_metres` was `1`, `rot_x_deg` was `0`, and `rot_y_deg` was `45`, the marker would be 45 degrees to the right of the robot, exactly 1 metre away.
+{{% /notice %}}
```
PeterJCLaw: :+1:
For even more clarity, could we get a diagram here? (Maybe out of scope for this PR though)

content/api/vision/coordinates.md
```diff
@@ -0,0 +1,34 @@
...
+---
+
+[Polar coordinates](https://en.wikipedia.org/wiki/Polar_coordinate_system) are represented by a distance from a set point and an angle from a set direction. [Cartesian coordinates](https://en.wikipedia.org/wiki/Cartesian_coordinate_system) are ones that are defined using a unique set of values (typically _x_, _y_ and _z_) to specify a point in space.
+
+{{% notice note %}}
+All distances are measured in metres.
+{{% /notice %}}
+
+## Polar Coordinates
+The polar coordinates are taken from the position of the webcam.
```
PeterJCLaw: I'm not sure this sentence reads as we want it to. How about phrasing this more like the Cartesian coordinates description is; something like "The polar coordinates of a marker are relative to the view of the camera."?

content/api/vision/marker-ids.md
```diff
@@ -0,0 +1,83 @@
...
+There are some useful constants which can be imported to do this. For example:
+
+```python
+from robot import WALL, TOKEN_ZONE_1, COLUMN_N_FACING_N
+```
+
+To check your marker matches one of these types, you can use the `in` operator:
+
+```python
+marker = r.camera.see()[0]  # Only the closest marker
```
PeterJCLaw: Could we introduce the tuple unpacking spelling of `marker, * = r.camera.see()` rather than indexing?

content/api/vision/marker-ids.md
```diff
@@ -0,0 +1,83 @@
...
+
+marker.id in WALL
+>>> True 
+```
+
+To see the full range of constants, and the exact IDs they match up to, have a look at [the source](https://github.com/sourcebots/robot-api/blob/master/robot/game_specific.py).
```
PeterJCLaw: Can we please _not_ do this? Asking people (especially learners) to look at the source code is not very friendly.

PeterJCLaw: For clarity: I think we should list all the constants in the docs, even if we don't have explanations for them all.

content/api/vision/marker-ids.md
```diff
@@ -0,0 +1,83 @@
...
+
+### _Wall_ Markers
+These markers are the ones positioned around the arena, there are 28 in total. Each token has a unique ID, and its location is documented in [the rules](/rules). This may be useful to calculate your position in the arena.
+
+```python
+marker.id in WALL  # Is this marker a wall marker
+```
+
+{{% notice tip %}}
+It's recommended to use `marker.is_wall_marker()` instead of `marker.id in WALL`
```
PeterJCLaw: This feels backwards to me -- if this is the recommended way to do this, then we should prioritise documenting that. I'd even argue that if using the helper is the recommended way then we shouldn't document the constant at all. 

This somewhat follows the logic of "There should be one, and preferably only one, obvious way to do it". (Unfortunately taking that to its logical extreme suggests that the helper shouldn't be there at all and we should just use (only) the constants since that would be consistent and thus obvious.)

content/api/vision/_index.md
PeterJCLaw: I agree the computation time here is probably fairly minimal (since `N` is likely to be < 1000 in all cases). I'm most compelled here by the argument about making our APIs feel as much like real-world (good, well designed) APIs.

While I agree that this isn't strictly a documentation thing, if we document it as being the case then we're stuck with it. If we don't document it, then it's undefined behaviour that we can remove.

content/api/vision/marker-ids.md
```diff
@@ -0,0 +1,83 @@
...
+
+```python
+marker.id in COLUMN
+>>> True
+```
+
+Each column has its own constant to check if the marker belongs to a specific column. There's also additional constants to check if a marker is facing a certain direction, and of course checking a specific face of a specific column.
+
+{{% notice tip %}}
+Columns are referenced like the points of a compass (_N_, _E_, _S_, _W_), with the North West corner as corner 0.
```
PeterJCLaw: Maybe clarify the corners as being parts of the arena: "... North West corner of the arena as corner 0"?

content/api/vision/marker-ids.md
```diff
@@ -0,0 +1,83 @@
...
+
+marker.id in COLUMN_FACING_E  # Is the marker facing East on its column?
+ 
+marker.id == COLUMN_S_FACING_W  # Is the marker on the South column facing West?
+```
+
+In the above examples, the points of the compass may be replaced with any others. 
+
+{{% notice note %}}
+The constants for a specific column face (e.g. `COLUMN_S_FACING_W`) only contain a single ID, so `==` should be used over `in`.
```
PeterJCLaw: This comment is potentially misleading -- it's not the number of ids that matters but the type of the value. 

Consider a 1-tuple (such as `foo = (0,)`) -- this contains only a single id, yet you'd still need to use `in` for a valid comparison to a marker id.

Perhaps say "The constants for the general classes of column markers (e.g: ...) are collections of marker ids and so need comparison with `in`. The constants for the specific column faces (e.g: ...) define only a single value and so should be compared to using `==`"?

content/api/vision/coordinates.md
RealOrangeOne: I like the looks of that, however I think it's worth changing the bounds, as i'm almost 100% sure the camera does not have a 180 degree FoV, which might lead to confusion

content/api/vision/coordinates.md
RealOrangeOne: To me, _"The y axis runs vertically through the camera"_ sounds wrong?

content/api/vision/marker-ids.md
RealOrangeOne: Does that work for lists too?

I'm not sure I like attempting to introduce students to this _advanced_ syntax like this, as they may not understand what it's doing

content/api/vision/marker-ids.md
RealOrangeOne: I'm going to leave this in for now. I agree that removing the helper function is the nicest approach, but that probably warrants further discussion

kierdavis: :+1: for deprecating `is_wall_marker` for consistency.

Carded up [here](https://trello.com/c/tHwRNdfo/63-deprecate-remove-iswallmarker-istokenmarker-methods).

content/api/vision/marker-ids.md
```diff
@@ -0,0 +1,138 @@
...
+marker.id in WALL  # Is this marker a wall marker
+```
+
+{{% notice tip %}}
+It's recommended to use `marker.is_wall_marker()` instead of `marker.id in WALL`
+{{% /notice %}}
+
+
+### _Column_ Markers
+These markers are those positioned on the 4 faces of the 4 columns in the arena. Each face has a unique, documented ID, and so could be used in addition to [_Wall_ Markers](#wall-markers) to calculate your position, and make sure you don't run into them!
```
kierdavis: "4" -> "four" (same reason as in 8b6a6268f0cbcbf4066f21ea66504b41f7d5a345)

content/api/vision/marker-ids.md
PeterJCLaw: It does indeed work for lists. (It turns out that you do need to name the `*` variable: `x, *y = [...]`, though that's unrelated to the unpacking -- the unpacking works for lists in Python 2).

I think my main objections here are:
- list indexing in python is generally a signal that you're doing something a bit wrong
- we're encouraging code which is intolerant of error cases
- we're reinforcing the notion (which as noted previously I don't like) that the markers are sorted

Rethinking this: since we're not defining the `r` here explicitly, I'm not actually sure why we feel a need to define the marker explicitly either. It could be viable to skip these lines entirely.

If we're merely trying to introduce the `in` operator here, then maybe we should back up to something where it's clearer what the operator is doing:
``` python
num = 4
num in [3,1,4,1,5]
>>> True
fibonacci = [1,1,2,3,5]
num in fibonacci
>>> False
```

content/api/vision/marker-ids.md
```diff
@@ -0,0 +1,138 @@
...
+
+marker.id in COLUMN_FACING_E  # Is the marker facing East on its column?
+ 
+marker.id == COLUMN_S_FACING_W  # Is the marker on the South column facing West?
+```
+
+In the above examples, the points of the compass may be replaced with any others. 
+
+{{% notice note %}}
+The constants for the general classes of column markers (e.g: `COLUMN_N`) are collections of marker ids and so need comparison with `in`. The constants for the specific column faces (e.g: `COLUMN _S_FACING_W`) define only a single value and so should be compared to using `==`.
```
kierdavis: There's a stray space in `COLUMN_S_FACING_W` which should be removed.

content/api/vision/marker-ids.md
```diff
@@ -0,0 +1,138 @@
...
+#### Markers facing a specific direction
+A set of marker ids specifying which direction a column is facing.
+
+- `COLUMN_FACING_N`
+- `COLUMN_FACING_E`
+- `COLUMN_FACING_S`
+- `COLUMN_FACING_W`
+ 
+#### Specific column faces
+A single marker ids specifying which column the marker is on, and which direction it's facing.
```
kierdavis: "ids" -> "id" (or even better, "ID")

content/api/vision/marker-ids.md
PeterJCLaw: I feel quite strongly that we shouldn't introduce something and then immediately say that something else is the recommended way to do things. I'm happy to postpone actually changing the API, but I think that we should resolve which is the recommended solution before documenting it.

content/api/vision/marker-ids.md
RealOrangeOne: Not sure i'll ever use id/ID correctly, however it's meant to be? (cc @PeterJCLaw )

Agreed it should be singular though

content/api/vision/marker-ids.md
RealOrangeOne: I'll drop this recommendation, and we can discuss dropping the helper in the card

content/api/vision/marker-ids.md
PeterJCLaw: I think lowercase is probably right (it's not an acronym), but I don't feel strongly. Agree it should be singular.

PeterJCLaw: Hrm, in some scenarios it could be an acronym for "Identity Document", though I think here it's just a contraction of "identity".

content/api/vision/marker-ids.md
RealOrangeOne: There's a note on the api intro that states `r` is implicitly defined as a `Robot` instance, http://docs.sourcebots.co.uk/api/#setup.

I don't think it's necessary to define _what_ the `in` operator does (the whirlwind tour should handle that), but I think keeping it to known, simple syntax like that is totally fine here. 

I'm not exactly sure what the issue is with indexing is, however maybe using `.pop()` is more explicit?

content/api/vision/marker-ids.md
RealOrangeOne: According to [the internet](https://stackoverflow.com/questions/1151338/id-or-id-on-user-interface), _ID_ is fine, however it should behave as _Id_ in camel-case environments. Personally, i'm ok with this! Although changing it might be out of scope for this PR

content/api/vision/marker-ids.md
PeterJCLaw: Essentially indexing usually doesn't match up with what you're trying to do and there are frequently  better python idioms to use that would match up.

Normally you'd want to loop over all the markers to find the one you're looking for, so you wouldn't index into the list. I'd rather not encourage indexing as it leads to this sort of thing:
``` python
for i in range(len(markers)):
    marker = markers[i]
```

I'm not sure how to avoid some form of element access here, hence the suggestion to just avoid it. (`.pop()` doesn't avoid that and does introduce issues like modifying the list).

content/api/vision/marker-ids.md
```diff
@@ -0,0 +1,133 @@
...
+There are some useful constants which can be imported to do this. For example:
+
+```python
+from robot import WALL, TOKEN_ZONE_1, COLUMN_N_FACING_N
+```
+
+To check your marker matches one of these types, you can use the `in` operator:
+
+```python
+marker, * = r.camera.see()  # Get only the closest marker
```
PeterJCLaw: Please can we drop the "closest marker" bit here? While I agree that it's correct, it's irrelevant.

PeterJCLaw: Note that this is currently invalid (it will `SyntaxError`) as you need to name the variable for the rest of the items.

How about:
``` python
first_marker, *rest = r.camera.see()
```
(which also drops the need for the comment too)

PeterJCLaw: (In retrospect I agree that introducing tuple unpacking here may not be the best thing either).

RealOrangeOne: I think explaining what the syntax is producing helps explain what's going on. I could live with changing 'closest' to 'first' perhaps?

RealOrangeOne: I'm going to revert and leave the syntax with indexing. Any benefits of trying to not tell people to index are outweighed by potentially odd looking code that's not directly addressed.

PeterJCLaw:
@RealOrangeOne could you resolve the conflicts before I re-review?

content/api/vision/coordinates.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Polar Coordinates
+The polar coordinates of a marker are relative to the view of the camera.
+
+ - `distance_metres` - the distance from the camera to the marker.
+ - `rot_x_rad` / `rot_x_deg` - the angle of rotation in radians/degrees about the x axis. The x axis runs horizontally in front of the camera. A positive angle would be one where the marker is in front of the camera.
+ - `rot_y_rad` / `rot_y_deg` - the angle of rotation in radians/degrees about the y axis. The y axis runs vertically in front of the camera. A positive angle would be one where the marker is rotated clockwise about the y axis looking from the top of the marker.
+ 
+{{% notice info %}}
+For example, if `distance_metres` was `1`, `rot_x_deg` was `0`, and `rot_y_deg` was `45`, the marker would be 45 degrees to the right of the robot, exactly 1 metre away.
```
PeterJCLaw: This example matches my expectation of the polar coordinates (and what I've implemented in https://github.com/sourcebots/sb-vision/pull/23), however is inconsistent with what's stated above.

In the above system, the `rot_x` value here makes the `rot_y` value meaningless as 0° is described as being above the robot.

Alternatively in the above system, the description here would instead be `Polar(rot_x_deg=90, rot_y_deg=45, distance_meters=1)`.

content/api/vision/coordinates.md
```diff
@@ -0,0 +1,34 @@
...
+{{% /notice %}}
+
+
+## Cartesian Coordinates
+The Cartesian position is represented as 3 coordinates, `x`, `y` and `z`. These coordinates are relative to the view of the webcam.  
+
+- `x` - the x coordinate, positive in the right direction.
+- `y` - the y coordinate, positive in the up direction.
+- `z` - the z coordinate, defined as the distance forwards in front of the camera.
```
PeterJCLaw: Minor: I think this would read better if we dropped "defined as".

content/api/vision/coordinates.md
RealOrangeOne: I think more testing (or code reading) is required to work out exactly what value of `rot_y` is directly infront of the robot at webcam level

PeterJCLaw: I don't think that testing is the right approach -- this is something we can define; we just need to work out what we want it to be. I'll have another read of the current polar implementation, though my current understanding is that it neither matches any of these nor even makes sense.

Having read https://en.wikipedia.org/wiki/Spherical_coordinate_system some more, It looks like we're unfortunately not using a standard anyway; we should have `rot_x` (_approximately_ θ in the ISO/physics model) and `rot_z` (φ) if we were. However even that is inaccurate since `rot_x` diverges from `θ` for large values of `rrot_z`.

If we change this at all, we're technically making a breaking change. I'm therefore tempted to suggest that we should break as _loudly_ as possible if we do need to break things. I see the following options:
 - drop polar coordinates entirely for now. They're not hard to work out, but they are hard to describe/document.
 - adopt the physics/ISO convention and change the property names (also rename to `spherical` rather than `polar`)
 - make no changes to the functionality, just document what's there
 - make no changes to the functionality, but remove the polar docs

For this PR I'm most tempted to just remove the polar docs, so that we get the other tidyups sooner, and then come back to this later.

PeterJCLaw: Ok, so following a little investigation, I've created some diagrams which capture my understanding of the current state of things. (I've got them as SVGs so we could even adapt them to be useful in the docs if we wanted too).

### Current implementation
Here the `polar_x` and `polar_y` values to match the output from the current [`cartesian_to_polar`](https://github.com/sourcebots/sb-vision/blob/57675c144b89eedb7ef451483c5a69412bc44d4a/sb_vision/tokens.py#L161); they translate to `rot_x...` and `rot_y...` respectively on the `robot-api` side.

These angles are not really in the right pace to be used (they're the wrong side of the triangle for one), though they also express a "rotation along the {x,y} axis" rather than "rotation about the {x,y} axis". Essentially I think that `polar_x` was intending to be "What's the angle of turn equivalent to the Cartesian 'x' value?"

vision-reference-current-polar-implementation

### Docs inline description
The angles here are angles _about_ an axis, which makes more sense than the current implementation.

This measures `rot_x` down from vertical, rather than up from horizontal.

vision-reference-docs-inline

### Docs example
The angles here are (again) angles _about_ an axis, which makes more sense than the current implementation.

This measures `rot_x` up from horizontal, rather than down from vertical, and feels more natural to me.

vision-reference-docs-example

### Others
I've not done a diagram for the ISO spherical system as it doesn't decompose into two flat images well; the diagram on wikipedia is also pretty good.

RealOrangeOne: To me, the _Docs example_ setup makes the most sense. I think there's much room for improvement here, especially in making sure that what's implemented actually makes sense, and that it matches what's in the docs. 

These images are really nice, I think it might be better to use those over made-up examples to better explain what each measurement is of!

RealOrangeOne:
RE https://github.com/sourcebots/docs/pull/67#discussion_r165590799. I think there's a lot of work that _could_ be done on the coordinate system, and they should be done sooner rather than later. But I think they should be done outside this PR, and properly discussed, as this PR is already very large and noisy

content/api/vision/coordinates.md
PeterJCLaw: > These images are really nice, I think it might be better to use those over made-up examples to better explain what each measurement is of!

Thanks! I think it's worth keeping the written example as well -- this is a very complicated topic, so having lots of ways it's explained/demonstrated is a good thing.

------

The more I think about this I feel we shouldn't change the existing behaviour for those who are using it. While it's clearly not the best, I think it would be usable. Hence if people have already started using it they'll likely have figured out how it behaves by testing.

I suggest that we:
1. drop the docs for the `polar` behaviour entirely (in this PR); maybe mention that it exists but is broken
2. (separately) introduce a `spherical` attribute and document that, with documentation largely lifted from this PR

This avoids breaking anyone already using the `polar` support, while allowing us to provide a better solution.

Thoughts?

RealOrangeOne: Due to the fact vision is almost certainly broken (https://github.com/sourcebots/sb-vision/pull/22) I personally think it's perfectly ok to drop `polar` entirely. I'd rather fully deprecate an api (perhaps with a print and a docs link), than silently replace it and remove its docs. We just need to get the `spherical` system in.

PeterJCLaw: (context: since my above comment I've since done most of the work for my proposed approach; I'm open to abandoning it if we feel that if another approach is better though)

@RealOrangeOne note that at the moment the vision system is using the `'c270'` calibration data, so it should be wrong but somewhat consistently so -- we never actually shipped the tecknet support.

content/api/vision/coordinates.md
PeterJCLaw: This happens in #91.

PR #70, opened 2017-12-19
Clarify notice about reserved servo assembly pins

kierdavis:
To be precise, it's the Arduino that communicates with the Pi
using digital pins 0 and 1, not the servo shield as this sentence
previously implied.

This commit also slightly improves the wording of this sentence.


kierdavis:
Thanks :+1:

PR #71, opened 2017-12-19
add PDFs of markers

Adimote:
A team requested some markers on the tech day, and I've since got an email from them asking where it is. I suggest we push this through.


Adimote:
I've already checked with an app that the IDs and orientations of the markers are correct

content/api/vision.md
```diff
@@ -72,3 +72,14 @@ There are three main marker types in the game `WALL`, `TOKEN` and `TOKEN_ZONE_x`
   - `TOKEN_ZONE_1` - Markers in the 1st zone with IDs from 49 - 53
   - `TOKEN_ZONE_2` - Markers in the 2nd zone with IDs from 54 - 58
   - `TOKEN_ZONE_3` - Markers in the 3rd zone with IDs from 59 - 63
+
+## Marker PDFs
+
+You can download PDF files of all markers:
+
+- [Large markers][large-marker-pdf], (should be printed on A3)
```
RealOrangeOne: I suggest we actually tell them what tokens are in which file (by type not ID)

Adimote:
I think we should ship this now then push an update with better markers later. as a team emailed me this morning asking for a PDF, and I don't want to delay them

RealOrangeOne:
I disagree, i suggest we ship something that fulfils their needs rather than a shoddy solution quicker. Adding these changes shouldnt take long, but will lead to a fix we don't have to update later on!

kierdavis:
There's nothing stopping us from replying to the email now (which we should do ASAP, really) in order to satisfy a team's immediate need. In the meantime we should make this page as useful as is feasible before we publish it.

content/api/vision.md
kierdavis: I agree, would be good to say "wall" / "column" / "token" markers here, as that's how we refer to them elsewhere in the docs.

The commas in this two lines also seem unnecessary to me.

PR #75, opened 2017-12-22
Add more kit pictures

Adimote:
Add a picture of a pi, along with replacing the servo board picture with one in the same format as the others


kierdavis:
It looks like this includes parts of another PR (the marker PDFs one). Please could you rebase?

Also, if you're replacing the servo board photo, could you remove the old one as well?

RealOrangeOne:
Agreed with the rebase, it's bad to try and review at this stage. 

Also looks like the servo assembly image isnt used. I'd suggest removing the image I had of the assembly and replace it with this one.

Can we also rotate those images 180 degrees? To me it feels as if theyre upside down, ie ports should be at the top.

Adimote:
Just thought about copyright issues here, will find a different source for the pi image

RealOrangeOne:
Everything looks ok, although I agree with @kierdavis, the images look a little weird, especially as the Pi image now has lots of the writing upside down. Can we restore the previous rotation (No, i cant make up my mind on most things)

PR #76, opened 2017-12-29
Pedantry

kierdavis:
First up, apologies for the monolithic PR, but because most of these commits hit the same lines it'd probably be worse if it was split into multiple PRs.

This is the results of poring over the whole docs site and aiming to improve its

* clarity
* consistency
* technical accuracy
* presentation
* tidiness

I've grouped related changes into individual commits, so it's probably easier to review this on a commit-by-commit basis.


.spelling
```diff
@@ -1,48 +1,42 @@
...
+pycharm
 python3
-dropdown
-3.81mm
-camcons
-micro-usb
-Smallpeice
-LiPo
+scipy
+sourcebots
```
RealOrangeOne: The spellings file is case sensetive. If an entry isnt in lower case, then that case must be enforced in the doc. I suggest forcing _SourceBots_ casing, and probably on a few others too

content/api/_index.md
```diff
@@ -6,42 +6,42 @@ weight: 1
...
 
-## Serial Number
+## Serial number
 All kit boards have a serial number, unique to that specific board, which can be read using the `serial` property:
 
 ```python
 r.power_board.serial
 >>> 'P1'
-r.serial_board.serial
+r.servo_board.serial
```
RealOrangeOne: 🤦‍♂️

content/api/motor-board.md
```diff
@@ -16,22 +16,22 @@ When you have more than one Motor Board connected to your kit, they can be acces
...
 
 ```python
 r.motor_board.m0
 r.motor_board.m1
 ```
 
-## Powering Motors
-Motor power is controlled using PWM with 100% power being a duty cycle of 1. You set the power with a value between `-1` and `1` inclusive (where a negative value puts the motor in reverse).
+## Powering motors
+Motor power is controlled using [pulse-width modulation (PWM)][pwm]. You set the power with a fractional value between `-1` and `1` inclusive, where `1` is maximum speed in one direction, `-1` is maximum speed in the other direction and `0` sets the speed to zero.
```
RealOrangeOne: To match with the rest of the docs, can we keep the link href inline in the markdown

RealOrangeOne: > `0` sets the speed to zero

This isnt necessarily correct, or the wording should be changed at least. If the power is set to `0`, then the motors start braking, to me _setting the speed to `0`_ is closer to coasting?

content/api/power-board.md
```diff
@@ -2,14 +2,16 @@
...
-As there is only one power board, it is not accessed like a list like motors for example:
+The power board can be accessed using the `power_board` property of the `Robot` object.
 
 ```python
 main_board = r.power_board
 ```
 
-## Power Outputs
-The power boards 6 outputs can be turned on and off using the `power_on` and `power_off` functions respectively.
+As there is only one power board, it cannot be accessed from a list like motor boards can.
```
RealOrangeOne: Unfortunately, it can. https://github.com/sourcebots/robot-api/blob/master/robot/robot.py#L93

I think a good solution for this would be to remove the `power_boards` property on the robot. However that might lead to weirdness if multiple are plugged in, but teams won't need to worry about that

content/api/servo-assembly/GPIO.md
```diff
@@ -2,31 +2,31 @@
...
 ### `PinMode.INPUT`
-Allows reading the value from the pin
+In this mode, the digital state of the pin (whether it is high or low) can be read.
 
 ### `PinMode.INPUT_PULLUP`
-Allows reading the value from the pin using the built-in [pullup resistor](https://learn.sparkfun.com/tutorials/pull-up-resistors).
+Same as `PinMode.INPUT`, but with an internal [pull-up resistor](https://learn.sparkfun.com/tutorials/pull-up-resistors) enabled.
 
 ### `PinMode.OUTPUT_HIGH`
-Set the pin to high output. Equivalent to `PinValue.HIGH`
```
RealOrangeOne: I quite liked the idea of explicitly staging it's the same as the pin value. Any reason this was removed?

content/api/servo-assembly/servos.md
```diff
@@ -2,11 +2,11 @@
 title: Servos
 ---
 
-From the servo assembly, you can control the attached servos. By default, servos will be unpowered when your robot starts, and can freely rotate. Once you set a value, they then fix to that rotation. This also happens when you turn off your robot, or unplug your USB stick.
+You can control attached servo motors from the servo assembly. By default, servos will be unpowered when your robot starts, and can freely rotate when turned by hand. Upon setting a value, they will hold the corresponding position. They will become unpowered again when you turn off your robot, or unplug your USB stick.
```
RealOrangeOne: they also go unpowered when the position is set to `None`. Not sure if that needs to be mentioned here, or if it's ok to just be mentioned below?

content/api/vision.md
```diff
@@ -5,14 +5,14 @@ title: Vision
 Your webcam is how your robot will see markers.
```
RealOrangeOne: Many of the changes in this file are handled in #67, but i'll sort out conflicts there

content/api/vision.md
```diff
@@ -25,17 +25,17 @@ for marker in markers:
...
   - `id` - returns the id of the marker.
   - `pixel_centre` - returns the pixel coordinates of the centre of the marker.
-  - `polar` - returns a `PolarCoord` object containing the position of the marker in the Polar coordinate system.
+  - `polar` - returns a `PolarCoord` object containing the position of the marker in the polar coordinate system.
   - `cartesian` - returns a `CartCoord` object containing the position of the marker in the Cartesian coordinate system.
 
-## Cartesian and Polar Coordinate Systems
-The markers' position can be represented using both the Polar and Cartesian coordinate system. The two objects `CartCoord` and `PolarCoord` hold the position information for both of these systems.
+## Cartesian and polar coordinate systems
+The markers' position can be represented using both the polar and Cartesian coordinate system. The two objects `CartCoord` and `PolarCoord` hold the position information for both of these systems.
```
RealOrangeOne: If _polar_ is lower cased here, shouldn't _Cartesian_?

content/kit/power-board.md
```diff
@@ -2,36 +2,36 @@
...
 
-There is also a Micro USB B connector which should be used to connect the Brain Board for control of the power board.
+There is also a Micro USB B connector which should be used to connect the Raspberry Pi for control of the power board.
 
-Finally, there are connectors for external Start and On\|Off switches. You may connect any push-to-make latching switch for the On\|Off button, or a push-to-make button for the start button.
+Finally, there are connectors for external Start and On|Off switches. You may connect any latching switch for the On|Off switch, or a push-to-make button for the Start button.
 
 {{% notice note %}}
-If you intend to use only the internal On|Off switch, you must connect a CamCon to the On|Off connector with a wire connecting one pin to the other pin on the same connector.
+If you intend to use only the internal On|Off switch, you must plug a CamCon into the On|Off connector with a wire connecting one pin to the other pin on the same connector.
```
RealOrangeOne: It's probably worth stating here that the kit we provide them already has one attached

.spelling
kierdavis: Wasn't aware of that, thanks. I'll leave the case of spelling entries as they are.

content/api/motor-board.md
kierdavis: Hmm - personally I interpret "setting the speed to 0" to mean "the motor will try its best to make its speed 0", consistent with how it behaves when you set the speed to +1, -1, or any other value.

Either way, I'll clarify this by explicitly saying "braking" instead.

content/api/power-board.md
kierdavis: To be honest there's no reason why the kit couldn't support multiple power boards (so long as each one has its own battery), so I would advise against removing `power_boards`.

However, we don't want teams to try and use more than one (at least, not until we've verified it's safe to do so), so I'll remove the reference to `power_boards` from the docs.

content/api/servo-assembly/GPIO.md
kierdavis: I wasn't sure if it was intended for people to use a `PinValue` where one would normally use a `PinMode`. It would work anyway though, since both are correspond to the Python strings `"high"`/`"low"`.

If this isn't intended usage of the API, then this line made it sound like `PinMode.OUTPUT_HIGH` and `PinValue.HIGH` are interchangeable, when they shouldn't be considered to be so.

If this is intended usage, it seems like `PinValue` itself is redundant, seeing as both of its values are covered by `PinMode`.

content/api/servo-assembly/servos.md
kierdavis: Might as well add it here too.

content/api/vision.md
kierdavis: [_Cartesian_][c] is a proper noun (named after René Descartes), while [_polar_][p] isn't.

[c]: https://en.wikipedia.org/wiki/Cartesian_coordinate_system
[p]: https://en.wikipedia.org/wiki/Polar_coordinate_system

content/kit/power-board.md
kierdavis: Good idea, thanks.

kierdavis:
@RealOrangeOne Thanks for the review. I've addressed most of your comments.

content/api/servo-assembly/GPIO.md
RealOrangeOne: technically yes theyre used for roughly the same thing, I like the fact we can be distinct between a mode and a value

content/api/_index.md
```diff
@@ -6,42 +6,42 @@ weight: 1
...
 ```python
 from robot import Robot
 
 r = Robot()
 ```
 
-This will setup your robot and allow it to send and receive data from the robot's various boards. Once this has been setup, this object can be used for most of the functions of the robot.
+This will initialise your robot and allow it to communicate with the robot's various boards. Once this has been done, this `Robot` object can be used for most of the functions of the robot.
 
 The remainder of the tutorials pages will assume your robot is defined as `r`.
```
PeterJCLaw: This sentence uses a rather different meaning of "your robot" than we do elsewhere. It would be good to fix that while we're here.

content/api/_index.md
```diff
@@ -6,42 +6,42 @@ weight: 1
...
-Your code needs to be put on a USB drive in a file called `main.py`. On insertion, this file will be executed. The file is directly executed off your USB drive, with your drive as the working directory.
+Your code needs to be put on a USB drive in a file called `main.py`. On insertion into the robot, this file will be executed. The file is directly executed off your USB drive, with your drive as the working directory.
 
 {{% notice warning %}}
 If this file is missing or incorrectly named, your robot won't do anything. No log file will be created.
 {{% /notice %}}
 
 ## Start Button
-So your robot doesn't start moving as soon as the robot is finished setting up, the code is paused after setup until the _Start Button_ on the power board is pressed. There is a green LED next to the start button when the robot is finished setting up and the start button can be pressed.
+After the robot has finished starting up, it will wait for the _Start Button_ on the power board to be pressed before running your code, so that you can control when it starts moving. There is a green LED next to the start button which flashes when the robot is finished setting up and the start button can be pressed.
```
PeterJCLaw: Is this strictly true? The definition of "your code" as implied above is anything in the `main.py`. Surely it's actually paused inside the call to `Robot()`?

content/api/_index.md
```diff
@@ -6,42 +6,42 @@ weight: 1
...
 
 ## Logs
-So you can see what your robot did, what it didn't do, and any errors it raised, a log file is saved on the USB drive. The file is saved to `log.txt` in the root of the USB.
+A log file is saved on the USB drive so you can see what your robot did, what it didn't do, and any errors it raised. The file is saved to `log.txt` in the top-level directory of the USB drive.
 
 {{% notice warning %}}
-The log file is deleted on each run, so save it (Ideally off the USB) if you need to keep hold of it!
+The log file is deleted on each run, so copy it elsewhere if you need to keep hold of it!
```
PeterJCLaw: maybe s/deleted/replaced/ or say "The previous log file..."?

content/api/editor-support.md
```diff
@@ -4,26 +4,26 @@ title: Editor support
...
 
-Installing these definitions is optional, however will be incredibly useful if your editor supports this. All editors in our [recommended list](/tutorials/editors/) besides IDLE will support it.
+Installing these definitions is optional, but will be incredibly useful if your editor supports this. All editors in our [recommended list](/tutorials/editors/) besides IDLE will support it.
 
 ## Installation
-Installing our API, the code you interact with, can be installed using `pip`, Python's built-in package manager. `pip` should be installed on your machine, if it's not, you'll need to look up how to install it. See [here](https://stackoverflow.com/a/12476379) for troubleshooting on Windows.
+Our API, the code you interact with, can be installed using Python's built-in package manager `pip`. Assuming you have Python installed on your machine, `pip` should be installed as well. If it's not, you'll need to look up how to install it. See [here](https://stackoverflow.com/a/12476379) for troubleshooting on Windows.
```
PeterJCLaw: `pip` is _not_ installed by default on Ubuntu... though fixing that feels out of scope for this PR.

content/api/editor-support.md
```diff
@@ -4,26 +4,26 @@ title: Editor support
...
 
 ```bash
 pip install git+https://github.com/sourcebots/robot-api
 ```
 
-If you're doing this on a shared computer, such as those you may find in your colleges, add `--user` to the end:
+The above command requires administrator access to the computer. If you're doing this on a shared computer, such as those you may find in your colleges, add `--user` to the end:
```
PeterJCLaw: `pip install` doesn't need administrator access (at least not in all configurations). This change also feels like it should have been a separate PR.

content/api/motor-board.md
```diff
@@ -16,22 +16,22 @@ When you have more than one Motor Board connected to your kit, they can be acces
 motor_board = r.motor_boards['SERIAL']
 ```
 
-This board object has a property for each of the motors connected to it. These are mapped to the `m1` and `m0` variables. The boards are labelled so you know which motor is which.
+This board object has a property for each of the motors connected to it. These are mapped to the `m0` and `m1` variables. The boards are labelled so you know which motor is which.
```
PeterJCLaw: While we're here can we also change "variables" to "attributes" (or maybe "properties") which is more correct.

content/api/motor-board.md
```diff
@@ -41,19 +41,19 @@ r.motor_board.m1
...
-Setting an incorrect value will raise an exception and your robot will crash.
-{{% /notice %}}  
+Setting a value outside of the range `-1` to `1` will raise an exception and your code will crash.
+{{% /notice %}}
 
-### Special Values
+### Special values
 
-In addition to the numeric values, there are two text constants that can be used. `robot.BRAKE` and `robot.COAST`. 
+In addition to the numeric values, there are two special constants that can be used: `robot.BRAKE` and `robot.COAST`.
```
PeterJCLaw: Another point which is probably for another PR: I don't like that we use `robot.BRAKE` spelling in text and then use it unqualified in the code example (`r.motor_board.m0 = BRAKE`). I'm not sure what the fix is.

content/api/motor-board.md
```diff
@@ -72,6 +72,5 @@ r.motor_board.m1 = COAST
 ```
 
 {{% notice warning %}}
-Changing the motor speed by more than 1 in a single operation (e.g. `-1` to `0`, `1` to `-1` etc) will likely trigger the current protection and your robot will shut down with a distinct beeping noise.
-{{% /notice %}} 
-
+Changing the motor speed by more than 1 in a single operation (e.g. `-1` to `0`, `1` to `-1` etc.) will likely trigger the current protection and your robot will shut down with a distinct beeping noise.
```
PeterJCLaw: The first example change (`-1` to `0`) seems wrong -- could we either explain why it's right or fix it?

content/api/motor-board.md
PeterJCLaw: One fix would be to use an enum for this rather than constants. (We can keep the constants for backwards compatibility and just drop them from the docs). `PinMode` shows well how this can work and be documented.

content/api/servo-assembly/GPIO.md
```diff
@@ -37,9 +37,9 @@ from robot import PinMode
 r.servo_board.gpios[3].mode = PinMode.INPUT_PULLUP
 ```
 
-## Reading a pin value
+## Reading a pin state
 
-To read the value of a pin, simply call `read()`. This will return either `PinValue.HIGH` or `PinValue.LOW`, which have to be imported to use.
+To read the state of a pin, simply call `read()`, which will return either `PinValue.HIGH` or `PinValue.LOW`. Like `PinMode`, the `PinValue` enum must be imported.
```
PeterJCLaw: I think we should keep (or add) the phrasing "imported to be used". Otherwise it could be read that calling `.read()` without the import would do something undefined.

content/api/servo-assembly/ultrasound.md
```diff
@@ -2,17 +2,16 @@
 title: Ultrasound
 ---
 
-We noticed many teams used ultrasound sensors to assist the vision system for finer distance measurement. So we've built in the mechanisms for reading and measuring distances using ultrasound right into the kit!
+We noticed many teams used ultrasound sensors to assist the vision system for finer distance measurement, so we've built the mechanisms for reading and measuring distances using ultrasound right into the kit!
```
PeterJCLaw: Are we also providing ultrasound hardware? What happens if the competitors provide their own {additional,replacement,etc.}?

content/api/vision.md
```diff
@@ -5,14 +5,14 @@ title: Vision
 Your webcam is how your robot will see markers.
 
 ## Using the camera to search for markers
-If you have a webcam connected, you can use `r.camera.see()` to take a picture. 
+If you have a webcam connected, you can use `r.camera.see()` to take a picture.
 
 If you consider it necessary you can assign the camera object to a variable just like `Robot`:
```
PeterJCLaw: This sentence is amazingly misleading (in terms of what `Robot` is as compared to `r.camera`). Can we just drop it and the following code snippet? (maybe a separate PR)

content/api/editor-support.md
RealOrangeOne: It does when installing globally right? Modifying the system environment like that usually requires admin privileges

content/api/_index.md
kierdavis: Good spot - this sentence seems to be more or less a duplicate of the sentence preceding the code sample anyway.

kierdavis: Oops, I mistakenly thought you were referring to the line changed in the diff rather than the line this thread is attached to. I've cleaned both up anyway.

content/api/_index.md
kierdavis: Changed "running your code" to "continuing with your code".

content/api/editor-support.md
kierdavis: Oops, wasn't really thinking about the meaning of this paragraph when I reworded it. Have added a mention that this is Windows-specific behaviour.

content/api/motor-board.md
kierdavis: I'll change to "attributes" rather than "properties", since while both of these terms could be used to describe `m0` and `m1` "attribute" is less tied to the implementation details.

content/api/motor-board.md
kierdavis: I agree that our usage of `robot.BRAKE` vs `BRAKE` appears inconsistent. There's also potential for confusion amongst newcomers between the class `Robot` and the module `robot` - especially since all our examples import names from the module (`from robot import X`) rather than importing it directly (`import robot`).

For now I think ensuring we demonstrate `from robot import BRAKE, COAST` and use the unqualified form `BRAKE` exclusively should be sufficient. I don't think an enum would be necessary here, in part because I'm not sure how one would name and describe it in a way that wouldn't also lend itself to including other motor modes (i.e. not brake or coast).

content/api/motor-board.md
kierdavis: Please could you clarify what you mean by this? Are you referring to the fact that the difference between `-1` and `0` is not strictly "more than 1"?

content/api/servo-assembly/GPIO.md
kierdavis: Agreed, done.

content/api/servo-assembly/ultrasound.md
kierdavis: To be honest, I think this whole paragraph feels out of place. We don't write like this for any other features of the kit such as servo control or vision, which makes it feel somewhat like another case of SR-vs-SB, something I'd like us to avoid.

I don't think we'd lose anything by removing this paragraph and replacing it with something along the lines of "The kit supports any ultrasonic sensor module compatible with the HC-SR04 standard."

Thoughts?

kierdavis: (this is probably worthy of separate PR)

content/api/vision.md
kierdavis: I've removed the unnecessary reference to `Robot` in 14e4544802301cb35f6df2c833908349bbbe6635, but left the rest as is.

Separately, I feel this block (and the corresponding ones for the power, motor and servo boards) don't add anything to the docs as they are. Users who are competent enough to structure their code in a way that lends itself to only passing around a single board rather than the entire `Robot` instance probably grasp the concept of attribute access already. For those who don't this snippet may just add to the confusion, as we don't clearly explain what the _purpose_ of doing this is or in what circumstances it's _necessary_.

I'd recommend we either try and make them much more useful, or just remove them on the basis of them being confusing. Either way this would be a separate PR.

content/api/editor-support.md
PeterJCLaw: I'm not sure. I've got it installed it via `chocolatey` so I'm not actually sure what options were set when it installed, though the Python directory is in the location which I expect when installing it "globally":
```
C:\Users\Peter>where pip
C:\Python36\Scripts\pip.exe
C:\Python27\Scripts\pip.exe
```

I think we want to say something like "Th above command _may_ require administrator.."

content/api/motor-board.md
PeterJCLaw: Yes: the difference between -1 and 0 is 1, which is not "more than 1".

content/api/servo-assembly/ultrasound.md
PeterJCLaw: Yeah, your suggestion is a much better way of presenting this; even better if we can link to the standard too.

content/api/editor-support.md
RealOrangeOne: I've got it installed using the installers on python.org. It installs to a different location to yours, howerver doesn't seem to require admin. On my linux machine however, I did need root or `--user`. I think perhaps moving this from inline in the instructions, and just adding a note about possibly needing `sudo`/`--user` in a notice or something would be the best solution. Certainly outside the scope of this PR

content/api/vision.md
PeterJCLaw: I _think_ it was introduced as we saw some SR competitors doing approximately the following (which doesn't work):
``` python
m0 = R.motors[0].m0
m0 = 50 # Set to 50% power
```

Since SourceBots' API has the potential for similar downfalls, I can see why theses are here; however I agree that they don't really feel like they add a lot in their current form.

content/tutorials/python-whirlwind-tour.md
```diff
@@ -520,11 +520,11 @@ else:
 
 Run the code and check your prediction.
 
-### Age Discrimination Tool
+### Age discrimination tool
```
PeterJCLaw: Maybe we shouldn't call it this? How about "Age detection tool" or similar?

content/api/editor-support.md
PeterJCLaw: I agree there's more that could be done; I think that adding a "may" to the current sentence improves it substantially (for all platforms) though.

kierdavis: I've added a "may" as suggested (b4838f2aac7d6f0583b547adfb2e058a545dc085), but this could certainly do with some more investigation.

content/api/motor-board.md
kierdavis: 6946af40a1a0d2e59ed791dd178bd7b74d89e60b

content/api/motor-board.md
kierdavis: I see. As far I can tell the current threshold of 1 is completely arbitrary, so how about we change this to "Large changes in the motor speed (e.g. `-1` to `0`, `1` to `-1` etc.)..."?

Perhaps we could recommend that the motor speed be changed only in small increments as well, so that users have examples of both "what is bad" and "what is good".

content/api/servo-assembly/ultrasound.md
kierdavis: Broken out into #77.

Unfortunately, it doesn't seem like there is an official standard for ultrasound sensors. It just looks like the vast majority of clones of this particular type of ultrasound module are compatible, and refer to themselves as "HC-SR04".

content/api/editor-support.md
RealOrangeOne: Have addressed in https://github.com/sourcebots/docs/pull/79

content/api/vision.md
RealOrangeOne: Addressed in https://github.com/sourcebots/docs/pull/67/commits/f969daf472fd3d76a49f083ace10d1353728ff94

content/tutorials/python-whirlwind-tour.md
kierdavis: 907c136

content/api/motor-board.md
kierdavis: addressed in 71d0b125be69ec2f51083e2da1c1ca907dffa9c4

content/api/motor-board.md
```diff
@@ -46,13 +46,17 @@ Setting a value outside of the range `-1` to `1` will raise an exception and you
 
 ### Special values
 
-In addition to the numeric values, there are two special constants that can be used: `robot.BRAKE` and `robot.COAST`.
+In addition to the numeric values, there are two special constants that can be used: `BRAKE` and `COAST`. In order to use these, they must be imported from the `robot` module like so:
```
PeterJCLaw: :+1:

kierdavis:
Thanks guys :+1:

PR #78, opened 2018-01-01
Add a Nix environment definition

kierdavis:
This allows users running the Nix package manager to instantly get
an environment containing `hugo` and `npm` by running `nix-shell shell.nix`.

Includes a version check for `hugo >= 0.30`. This is important because
the current stable release of the Nix package set (17.09) ships version 0.29,
which doesn't meet our minimum requirements. Unfortunately I haven't
managed to get this to print anything nicer than "assertion failed"
when the version check fails.

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -1,0 +1,14 @@
+{ pkgs ? import <nixpkgs> {} }:
+
+with pkgs;
+
+# Require hugo >= 0.30.
+assert (builtins.compareVersions hugo.version "0.30") >= 0;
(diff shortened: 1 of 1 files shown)


kierdavis:
Indeed it does :+1:

```
$ nix-build '<nixpkgs>' -A nodejs && ls result/bin
/nix/store/v2jlrnp667chyxsfyxk82xlgwnmvs625-nodejs-6.11.5
node  npm
```

PR #81, opened 2018-01-06
Specify npm is only for tests

RealOrangeOne:
`npm` is only needed for tests (specifically spell checker), so it's not necessary to have to build the site.

@kierdavis not sure if `nix` supports some kind of _test dependency_ entry?

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ URL: http://docs.sourcebots.co.uk
 
 ## Requirements
 - [Hugo](https://gohugo.io) (>=0.30)
-- [NodeJS](https://nodejs.org/) (>=6)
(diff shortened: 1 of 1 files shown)


kierdavis:
There isn't an explicit "test dependency" option, but typically this is done by adding an optional boolean argument and using that to control whether `nodejs` is pulled in or not. Like I've done [here](https://github.com/sourcebots/docs/pull/82/commits/4d93f6e82d8d2e9806a6fb285b499326bbbceb62) in #82.

kierdavis:
Separately, it might be a good idea to briefly mention what the tests do, so that users can decide whether they want to run them or not.

kierdavis:
:+1:

PR #87, opened 2018-01-31
Add documentation for custom arduino commands

PeterJCLaw:
This includes a summary of how to create the command on the Arduino as well as how to call it from Python and the errors which could be thrown.

See https://github.com/sourcebots/robot-api/pull/55 and https://github.com/sourcebots/robotd/pull/35 for the related implementations.

Note: there are some TODOs inline. The ones I think need fixing before we ship are:
 - [x] understanding (and documenting) what happens to values which are sent as numbers (either float or integer)
 - [x] what kind of Arduino is it?

### DIFF
--- a/content/api/servo-assembly/_index.md
+++ b/content/api/servo-assembly/_index.md
@@ -9,7 +9,7 @@ The Servo Assembly is used to connect to multiple types of peripheral:
 - [Ultrasound sensors](ultrasound)
 
 At the core of the Servo Assembly is an Arduino. If you need more direct control
-over your perhiperals, you can change the firmware running on the Arduino and
+over your peripherals, you can change the firmware running on the Arduino and
 extend it to support [custom commands](arduino-commands).
 
 ## Getting a servo board instance
--- a/content/api/servo-assembly/_index.md
+++ b/content/api/servo-assembly/_index.md
@@ -8,9 +8,9 @@ The Servo Assembly is used to connect to multiple types of peripheral:
 - [Servos](servos)
 - [Ultrasound sensors](ultrasound)
 
-At the core of the Servo Assembly is an Arduino <TODO: model name>. If you need
-more direct control over your perhiperals, you can change the firmware running
-on the Arduino and extend it to support [custom commands](arduino-commands).
+At the core of the Servo Assembly is an Arduino. If you need more direct control
+over your perhiperals, you can change the firmware running on the Arduino and
+extend it to support [custom commands](arduino-commands).
 
 ## Getting a servo board instance
 
--- a/content/api/servo-assembly/_index.md
+++ b/content/api/servo-assembly/_index.md
@@ -7,7 +7,10 @@ The Servo Assembly is used to connect to multiple types of peripheral:
 - [GPIO pins](gpio) (digital and analogue)
 - [Servos](servos)
 - [Ultrasound sensors](ultrasound)
-- [Custom Arduino commands](arduino-commands)
+
+At the core of the Servo Assembly is an Arduino <TODO: model name>. If you need
+more direct control over your perhiperals, you can change the firmware running
+on the Arduino and extend it to support [custom commands](arduino-commands).
 
 ## Getting a servo board instance
 
--- a/content/api/servo-assembly/_index.md
+++ b/content/api/servo-assembly/_index.md
@@ -7,6 +7,7 @@ The Servo Assembly is used to connect to multiple types of peripheral:
 - [GPIO pins](gpio) (digital and analogue)
 - [Servos](servos)
 - [Ultrasound sensors](ultrasound)
+- [Custom Arduino commands](arduino-commands)
 
 ## Getting a servo board instance
 
--- a/content/api/servo-assembly/arduino-commands.md
+++ b/content/api/servo-assembly/arduino-commands.md
@@ -15,12 +15,9 @@ than can be achieved through the Python API.
 
 As your new firmware will still need to be able to talk to the Python code
 running on the Raspberry Pi, you'll need to base it on SourceBots default
-firmware. You can get a copy of the source code from GitHub either by
-downloading the [repo archive][servo-firmware-zip] or by cloning the repo:
-
-```txt
-git clone https://github.com/sourcebots/servo-firmware
-```
+firmware. You can get a copy of the source code from its GitHub repo at
+<https://github.com/sourcebots/servo-firmware>. Note that you will also need the
+Adafruit PWM Servo Driver Library which is included in that repo as a submodule.
 
 We recommend using [Arduino Studio](https://www.arduino.cc/en/Main/Software) to
 edit your firmware as it provides direct support for building the code into the
@@ -210,5 +207,3 @@ correctly for all your messages back to the Pi.
     which the teams aren't using. That's intentionally omitted for the moment
     as it's nontrivial to describe which bits are safe to remove.
 -->
-
-[servo-firmware-zip]: https://github.com/sourcebots/servo-firmware/archive/master.zip
--- a/content/api/servo-assembly/arduino-commands.md
+++ b/content/api/servo-assembly/arduino-commands.md
@@ -97,8 +97,6 @@ use), you can have a look at the other commands the Arduino already supports.
 In particular the `get_version`, `led` and `servo` functions are a good starting
 point.
 
-<!-- TODO should we error in robot-api/robotd if the incoming arguments include spaces? -->
-
 ### Registering the command
 
 To register this command we need to modify the list of commands stored in the
--- a/content/api/servo-assembly/arduino-commands.md
+++ b/content/api/servo-assembly/arduino-commands.md
@@ -57,7 +57,7 @@ with the Pi, though you can otherwise do pretty much anything you normally would
 on an Arduino.
 
 The following example command shows how to handle arguments from the Pi as well
-as how to return errors an other values:
+as how to return errors and other values:
 
 ```cpp
 static CommandError echo(int commandId, String argument) {
--- a/content/api/servo-assembly/arduino-commands.md
+++ b/content/api/servo-assembly/arduino-commands.md
@@ -2,8 +2,8 @@
 title: Custom Arduino Commands
 ---
 
-At the core of the servo assembly is an Arduino <TODO: model name>. By default
-the Arduino runs firmware created by SourceBots providing simple access to the
+At the core of the servo assembly is an [Arduino Uno][arduino]. By default the
+Arduino runs firmware created by SourceBots providing simple access to the
 attached servo shield, compatible ultrasound sensors and to the GPIO pins
 directly.
 
@@ -193,5 +193,6 @@ correctly for all your messages back to the Pi.
     as it's nontrivial to describe which bits are safe to remove.
 -->
 
+[arduino]: https://store.arduino.cc/arduino-uno-rev3
 [servo-firmware-zip]: https://github.com/sourcebots/servo-firmware/archive/master.zip
 [errors]: https://docs.python.org/3/tutorial/errors.html
--- a/content/api/servo-assembly/arduino-commands.md
+++ b/content/api/servo-assembly/arduino-commands.md
@@ -54,7 +54,6 @@ static CommandError echo(int commandId, String argument) {
     return COMMAND_ERROR("nothing to echo");
   }
 
-  int i = 1;
   while (argument.length() > 0) {
     // You can use `pop_option` to extract individual parameters from the
     // `argument` string. They're sent space-separated, so be sure not to
--- a/content/api/servo-assembly/arduino-commands.md
+++ b/content/api/servo-assembly/arduino-commands.md
@@ -17,7 +17,7 @@ As your new firmware will still need to be able to talk to the Python code
 running on the Raspberry Pi, you'll need to base it on SourceBots default
 firmware. You can get a copy of the source code from GitHub:
 
-```none
+```txt
 git clone https://github.com/sourcebots/servo-firmware
 ```
 
--- a/content/api/servo-assembly/arduino-commands.md
+++ b/content/api/servo-assembly/arduino-commands.md
@@ -16,7 +16,7 @@ than can be achieved through the Python API.
 As your new firmware will still need to be able to talk to the Python code
 running on the Raspberry Pi, you'll need to base it on SourceBots default
 firmware. You can get a copy of the source code from GitHub either by
-downloading the [repo archive](servo-firmware-zip) or by cloning the repo:
+downloading the [repo archive][servo-firmware-zip] or by cloning the repo:
 
 ```txt
 git clone https://github.com/sourcebots/servo-firmware
--- a/content/api/servo-assembly/arduino-commands.md
+++ b/content/api/servo-assembly/arduino-commands.md
@@ -112,15 +112,15 @@ To call a command from Python you should call the `direct_command` method on a
 `Servo` instance. You can send strings and numbers:
 
 ```python
-r.servo_board.direct_command('echo', "first argument", 2, "third argument")
+r.servo_board.direct_command('echo', 'first argument', 2, 'third argument')
 ```
 
 This will return a `list` of the responses from the command (anything which was
 sent back using the `serialWrite` function):
 
 ```python
->>> r.servo_board.direct_command('echo', "first argument", 2, "third argument")
-["first argument", "2", "third argument"]
+>>> r.servo_board.direct_command('echo', 'first argument', 2, 'third argument')
+['first argument', '2', 'third argument']
 ```
 
 Note that the responses will all be strings, so you'll need to explicitly
--- a/content/api/servo-assembly/arduino-commands.md
+++ b/content/api/servo-assembly/arduino-commands.md
@@ -37,13 +37,32 @@ under the name you'll use to call it from the Python code.
 
 ### Implementing the command
 
-Here's a simple command which just echoes back any arguments it receives:
+You'll need to define a function to implement your command. The function should
+be placed in the `firmware.cpp` file just above the definition of the `commands`
+array (likely about like 205).
+
+The function definition needs to look like this:
 
 ```cpp
-// Define a function which implements your command. As long as you stick to
-// letters, numbers and underscore (_), you can call it whatever you like.
-// With the exception of the name (here it's `echo`), the rest of the function's
-// "signature" must be exactly as it is on the following line:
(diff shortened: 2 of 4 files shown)


content/api/servo-assembly/arduino-commands.md
```diff
@@ -0,0 +1,197 @@
...
+example if you have other types of sensor which require higher precision timing
+than can be achieved through the Python API.
+
+## Requirements
+
+As your new firmware will still need to be able to talk to the Python code
+running on the Raspberry Pi, you'll need to base it on SourceBots default
+firmware. You can get a copy of the source code from GitHub:
+
+```none
```
RealOrangeOne: Can we use `txt` here? I'm not even sure `none` is a valid highlight

content/api/servo-assembly/arduino-commands.md
```diff
@@ -0,0 +1,197 @@
...
+git clone https://github.com/sourcebots/servo-firmware
+```
+
+<TODO: describe how to do that without git>
```
RealOrangeOne: Can we use Go templates comment syntax rather than this or the html comments below.

`{{/* a comment */}}`

RealOrangeOne:
Could we perhaps change the name to _Custom arduino commands_, as it's more discoverable that way

content/api/servo-assembly/_index.md
```diff
@@ -7,6 +7,7 @@ The Servo Assembly is used to connect to multiple types of peripheral:
 - [GPIO pins](gpio) (digital and analogue)
 - [Servos](servos)
 - [Ultrasound sensors](ultrasound)
+- [Custom Arduino commands](arduino-commands)
```
RealOrangeOne: I'm not sure this link fits here, as this isnt a peripheral that's communicated with through the arduino.

content/api/servo-assembly/arduino-commands.md
PeterJCLaw: Aha, thanks. I searched hugo's docs for what the comment syntax was, but couldn't find it!

PeterJCLaw: 5328e1d resolves this TODO outright; I'll changes the other comments shortly.

content/api/servo-assembly/arduino-commands.md
PeterJCLaw: 79fbd5626c6fc9709c32fb47431a2fd8b270b661

content/api/servo-assembly/_index.md
PeterJCLaw: Good point; I'd not actually read the preceding line!

PeterJCLaw: 9784c1a129d5388d2c8f2b4883835bf00b765484

content/api/servo-assembly/arduino-commands.md
PeterJCLaw: Turns out that `{{/* a comment */}}` unfortunately doesn't work for me. I suspect it might be related to these being markdown sources, though it would be an odd bug if the comment stripper happened _after_ the markdown conversion.

content/api/servo-assembly/arduino-commands.md
```diff
@@ -0,0 +1,196 @@
...
+than can be achieved through the Python API.
+
+## Requirements
+
+As your new firmware will still need to be able to talk to the Python code
+running on the Raspberry Pi, you'll need to base it on SourceBots default
+firmware. You can get a copy of the source code from GitHub either by
+downloading the [repo archive](servo-firmware-zip) or by cloning the repo:
+
+```txt
```
kierdavis: Would `bash` (or similar) be a better choice than `txt` here?

content/api/servo-assembly/arduino-commands.md
```diff
@@ -0,0 +1,196 @@
...
+
+Adding a command requires two changes. First you must define the function which
+will implement the operations of that command and then you register that command
+under the name you'll use to call it from the Python code.
+
+### Implementing the command
+
+Here's a simple command which just echoes back any arguments it receives:
+
+```cpp
```
kierdavis: We should mention that this function definition must be placed before its use (the `commands` array), unless they add a prototype for the function as well of course.

content/api/servo-assembly/arduino-commands.md
```diff
@@ -0,0 +1,196 @@
...
+
+  // If everything went fine, you should return `OK`.
+  return OK;
+}
+```
+<!-- TODO: mention being able to send back comments? -->
+
+For other examples of how to define commands (and the helper functions you can
+use), you can have a look at the other commands the Arduino already supports.
+In particular the `get_version`, `led` and `servo`  are a good starting point.
```
kierdavis: Are you missing the word "commands" after "servo" here?

content/api/servo-assembly/arduino-commands.md
```diff
@@ -0,0 +1,196 @@
...
+static const CommandHandler commands[] = {
+  // ... lots of existing commands ... (make sure you leave them in place!)
+  CommandHandler("echo", &echo, "echoes its arguments, each on a new line"),
+};
+```
+
+The first argument is a quoted `String` literal containing the name you'll use
+in Python to call this command. It can contain letters, numbers and dashes (`-`).
+
+The second argument is the function you defined for your function, with an
```
kierdavis: >the function defined for your function

Should this say "the name defined for your function"? (IMO "the name of your function" would probably be more straightforward anyway)

content/api/servo-assembly/_index.md
```diff
@@ -8,6 +8,10 @@ The Servo Assembly is used to connect to multiple types of peripheral:
 - [Servos](servos)
 - [Ultrasound sensors](ultrasound)
 
+At the core of the Servo Assembly is an Arduino. If you need more direct control
```
kierdavis: Not a big deal, but is there any reason why "Arduino" here isn't hyperlinked in the same way that the occurrences in the introductions to `arduino-commands.md` and `kit/servo-assembly/_index.md` are?

content/api/servo-assembly/arduino-commands.md
```diff
@@ -0,0 +1,196 @@
...
+of the command. This helps document what the command is for, and will be
+returned as part of the message if you call the `help` command.
+
+## Calling commands
+
+To call a command from Python you should call the `direct_command` method on a
+`Servo` instance. You can send strings and numbers:
+
+```python
+r.servo_board.direct_command('echo', "first argument", 2, "third argument")
```
kierdavis: Any reason for the mixing of quote types here?

content/api/servo-assembly/arduino-commands.md
```diff
@@ -0,0 +1,196 @@
...
+This error will generally indicate that one of the messages being sent back from
+the Arduino is badly formatted. Check to ensure that you're using `serialWrite`
+correctly for all your messages back to the Pi.
+
+
+<!--
+    At some point we might need to detail how to remove bits of the firmware
+    which the teams aren't using. That's intentionally omitted for the moment
+    as it's nontrivial to describe which bits are safe to remove.
+-->
```
kierdavis: Haven't tested this but I would imagine that removing or commenting an entry in the `commands` away should cause the corresponding handler function to be optimised out of the binary. This should produce some initial space savings, at least.

content/api/servo-assembly/arduino-commands.md
```diff
@@ -0,0 +1,196 @@
...
+You can also extend the firmware on the Arduino to support more features, for
+example if you have other types of sensor which require higher precision timing
+than can be achieved through the Python API.
+
+## Requirements
+
+As your new firmware will still need to be able to talk to the Python code
+running on the Raspberry Pi, you'll need to base it on SourceBots default
+firmware. You can get a copy of the source code from GitHub either by
+downloading the [repo archive](servo-firmware-zip) or by cloning the repo:
```
kierdavis: Parentheses -> square brackets

Currently this is resulting in a broken link to [https://HOSTNAME/api/servo-assembly/arduino-commands/servo-firmware-zip](https://deploy-preview-87--sourcebots-docs.netlify.com/api/servo-assembly/arduino-commands/servo-firmware-zip).

content/api/servo-assembly/arduino-commands.md
PeterJCLaw: maybe, though that causes "clone" to be highlighted a different colour. I've no idea why -- as far as I know "clone" isn't a special word in any shell. I opted for it being less distracting even if it's slightly less correct.

content/api/servo-assembly/arduino-commands.md
PeterJCLaw: almost: 932d341

content/api/servo-assembly/arduino-commands.md
PeterJCLaw: I think I meant "the function defined for your command" (de16f254b40ccd768c5573d75213d5f86f438b49); do you think that that's clear enough that we mean the function as a symbol?

content/api/servo-assembly/_index.md
PeterJCLaw: Only that it didn't seem entirely useful to do so. Anyone wanting more detail will probably be on one of those pages anyway. Do we prefer to be over-linking?

content/api/servo-assembly/arduino-commands.md
PeterJCLaw: `'echo'` is for the machine and is close to being an enum value or a slug. The others are less so, though are arguably still much more that than my usual use of double quotes (which is for text intended for humans).

They should probably all be `'`.

content/api/servo-assembly/arduino-commands.md
PeterJCLaw: This does indeed seem to happen. I think I'd still rather defer this for now.

content/api/servo-assembly/arduino-commands.md
PeterJCLaw: good spot: 422a0f8

PeterJCLaw: Is there a way we can test that all internal links go somewhere?

content/api/servo-assembly/arduino-commands.md
PeterJCLaw: d1f728d

content/api/servo-assembly/arduino-commands.md
PeterJCLaw: b93eaf0 reworks the start of the "Implementing the command" section and includes this.

content/api/servo-assembly/arduino-commands.md
```diff
@@ -0,0 +1,216 @@
...
+You can also extend the firmware on the Arduino to support more features, for
+example if you have other types of sensor which require higher precision timing
+than can be achieved through the Python API.
+
+## Requirements
+
+As your new firmware will still need to be able to talk to the Python code
+running on the Raspberry Pi, you'll need to base it on SourceBots default
+firmware. You can get a copy of the source code from GitHub either by
+downloading the [repo archive][servo-firmware-zip] or by cloning the repo:
```
RealOrangeOne: Perhaps here, rather than specifically getting a zip or using `git clone`, we just link to the repo. This means they can not only see the source without downloading it, but the discovery path for clone / download instructions is very straightforward

content/api/servo-assembly/arduino-commands.md
```diff
@@ -0,0 +1,216 @@
...
+
+Adding a command requires two changes. First you must define the function which
+will implement the operations of that command and then you register that command
+under the name you'll use to call it from the Python code.
+
+### Implementing the command
+
+You'll need to define a function to implement your command. The function should
+be placed in the `firmware.cpp` file just above the definition of the `commands`
+array (likely about like 205).
```
RealOrangeOne: s/like/line/?

content/api/servo-assembly/arduino-commands.md
```diff
@@ -0,0 +1,216 @@
...
+correctly for all your messages back to the Pi.
+
+
+<!--
+    At some point we might need to detail how to remove bits of the firmware
+    which the teams aren't using. That's intentionally omitted for the moment
+    as it's nontrivial to describe which bits are safe to remove.
+-->
+
+[arduino]: https://store.arduino.cc/arduino-uno-rev3
```
RealOrangeOne: As we have done with almost all other links in the docs, can we use markdown links inline

content/api/servo-assembly/arduino-commands.md
RealOrangeOne: Turns out the actual reason for this not working is that Hugo doesnt run Go's templating engine over the content files, so never strips out the comments -_- D'oh!

content/api/servo-assembly/arduino-commands.md
PeterJCLaw: 3457afd

content/api/servo-assembly/arduino-commands.md
PeterJCLaw: For cases where that causes the source to wrap oddly I tend to prefer separate links. 83266fe however does this.

content/api/servo-assembly/arduino-commands.md
PeterJCLaw: e1dc43f

content/api/servo-assembly/arduino-commands.md
```diff
@@ -0,0 +1,209 @@
...
+
+Adding a command requires two changes. First you must define the function which
+will implement the operations of that command and then you register that command
+under the name you'll use to call it from the Python code.
+
+### Implementing the command
+
+You'll need to define a function to implement your command. The function should
+be placed in the `firmware.cpp` file just above the definition of the `commands`
+array (likely about line 205).
```
kierdavis: like -> line?

I feel like line numbers are quite a volatile way to refer to parts of the code, since changes to the file can result in many of the line numbers changing. IMO better solutions are to either copy the line verbatim here so users can `ctrl+f` for it, or just to say the rough location qualitatively (e.g. "about halfway through the file") - both of these methods are less likely to break when the file is modified.

content/api/servo-assembly/arduino-commands.md
kierdavis: hmm, weird. `txt` is fine here in that case.

content/api/servo-assembly/arduino-commands.md
kierdavis: We still say "name" later in the sentence, so I think that's clear enough :+1:

content/api/servo-assembly/_index.md
kierdavis: Fair enough, good point.

content/api/servo-assembly/arduino-commands.md
kierdavis: Yeah, I'd agree than in typical applications the arguments to the command won't be intended to be human-readable text - more likely to be numbers or keywords (e.g. 'low' / 'high').

content/api/servo-assembly/arduino-commands.md
kierdavis: Yeah, that's something for separate PR.

content/api/servo-assembly/arduino-commands.md
kierdavis: I'm sure a broken link checker (e.g. [broken-link-checker-local](https://www.npmjs.com/package/broken-link-checker-local)) would be a useful addition to the CI tests.

kierdavis: #89

content/api/servo-assembly/arduino-commands.md
```diff
@@ -37,13 +37,32 @@ under the name you'll use to call it from the Python code.
...
+`my_function`), though you'll need to stick to letters, numbers and underscore
+(`_`). The rest of the function signature (the line starting `static
+CommandError` above) needs to be exactly as above.
+
+You command will need to adhere to some conventions to be able to communicate
+with the Pi, though you can otherwise do pretty much anything you normally would
+on an Arduino.
+
+The following example command shows how to handle arguments from the Pi as well
+as how to return errors an other values:
```
kierdavis: "an" -> "and"

content/api/servo-assembly/arduino-commands.md
PeterJCLaw: I'd hope that this wouldn't change very much, also hence the "about" on the line number. Given that we do include the definition of the commands array below I don't want to repeat it here.

Do you think we need to add a comment like "see below" for this, or can we assume people will be able to read the whole of the documentation to get a full understanding?

content/api/servo-assembly/arduino-commands.md
PeterJCLaw: https://github.com/sourcebots/docs/commit/d59f768b2dfafa028290e976806422418439a093

content/api/servo-assembly/arduino-commands.md
kierdavis: I think we could just remove the remark - I think it should be easy enough to find `commands` in the source file.

content/api/servo-assembly/arduino-commands.md
```diff
@@ -0,0 +1,209 @@
...
+}
+```
+<!-- TODO: mention being able to send back comments? -->
+
+For other examples of how to define commands (and the helper functions you can
+use), you can have a look at the other commands the Arduino already supports.
+In particular the `get_version`, `led` and `servo` functions are a good starting
+point.
+
+<!-- TODO should we error in robot-api/robotd if the incoming arguments include spaces? -->
```
kierdavis: This isn't really related to the docs - could we break this out into an issue on `robot-api` or a Trello card instead?

PeterJCLaw: 2974f73

content/api/servo-assembly/arduino-commands.md
PeterJCLaw: I think it's useful to have the line number; I think it's clear enough that it's approximate.

content/api/servo-assembly/arduino-commands.md
```diff
@@ -0,0 +1,207 @@
...
+    // they are here; the third argument needs to be a `String` containing the
+    // value you want to send back to the Pi.
+    serialWrite(commandId, '>', next_arg);
+  }
+
+  // If everything went fine, you should return `OK`.
+  return OK;
+}
+```
+<!-- TODO: mention being able to send back comments? -->
```
Adimote: todo message here?

PR #92, opened 2018-02-02
Install broken link checker

### LINKED ISSUE #89: Add a broken hyperlink checker to the CI test phase
e.g. https://www.npmjs.com/package/broken-link-checker-local (though of course Google suggests many alternatives).

RealOrangeOne:
Fixes #89


RealOrangeOne:
Looks like #88 has caused some issues with URLs, as python doesnt store documents under patch versions. Will fix as part of this PR

kierdavis:
I can confirm that learnpython is behaving weirdly - it works in my browser (chromium) but gives a 500 when fetched through curl (on both HTTP and HTTPS). Perhaps it's trying to make some decision based on the `User-Agent` which just crashes when it see an unexpected value?

kierdavis:
I would recommend renaming `python_version` and `specific_python_version` to something more clear, such as `python_minor_version` and `python_patch_version`.

PR #93, opened 2018-02-03
Revert "Drop all references to polar co-ordiates"

RealOrangeOne:
This reverts commit 06ce18ad5ed64ed209215317a381f5da570ffe6e.

Because we haven't shipped the `spherical` code, or docs, we should still be documenting the `polar` coords, so teams can continue to use the kit.


kierdavis:
Context: there are teams at the ongoing tech day trying to find the documentation on `polar`, since that's what's shipped on their kits right now. This will be un-reverted (i.e. remove `polar` docs again) once `spherical` is shipped.

PeterJCLaw:
This is a terrible idea. The `polar` attribute as it stands is confusing at best and *does not match these docs*.

PR #94, opened 2018-02-04
Kit updating

RealOrangeOne:
Document the update mechanism for both flashing SD cards, and the update file

This PR doesnt contain the page to download files, that'll come later


content/kit/pi/update-file.md
```diff
@@ -0,0 +1,27 @@
...
+
+{{% notice tip %}}
+Using this file is the recommended way to update your kit.
+{{% /notice %}}
+
+## Using the update file
+
+1. Put the `update.tar.xz` file on a USB drive.
+2. Connect the USB drive to your robot.
+3. Turn on your robot
```
kierdavis: Full stops at the end of list items (see earlier comment).

content/kit/pi/sd-card.md
```diff
@@ -0,0 +1,26 @@
...
+
+{{% notice note %}}
+If you choose to use a tool other than Etcher, you may need to extract the `pi-image.img.xz` to `pi-image.img`. There are many tools available for this.
+{{% /notice %}}
+
+### Flashing
+
+1. Open Etcher and select the `pi-image.img.xz` file you downloaded.
+2. Select your SD card from the devices window
+3. Click 'Flash!'.
```
kierdavis: Could we have consistent use of full stops at the end of lines in this list? I'm happy with either all of them having or none of them having one, but not with only some. :)

content/kit/pi/_index.md
```diff
@@ -14,3 +14,13 @@ The software included on the Raspberry Pi allows it to run your robot. No graphi
 {{% notice warning %}}
 Please do not attempt to modify the software on the Raspberry Pi. It has been configured and tested in its current configuration. Unauthorized modifications may damage your robot or prevent accurate debugging.
 {{% /notice %}}
+
+## Updating your Pi
+Keeping your kit up to date is very important. It enables us to deploy new features, as well as fix bugs. 
+
+There are 2 ways of updating your kit:
+
+- [Using the `update.tar.xz` file](update-file)
```
PeterJCLaw: There is likely more than one update, so this should probably be "an" rather than "the".

content/kit/pi/update-file.md
```diff
@@ -0,0 +1,27 @@
...
+
+After the update is completed, the Pi will reboot. If you have a `main.py` file on the USB as well, this will now execute, because the `update.tar.xz` file is deleted when the update begins.
+
+{{% notice warning %}}
+Do not turn off your Pi during the update process. It can take a few minutes to finish.
```
PeterJCLaw: how will they know when the update is complete?

content/kit/pi/update-file.md
```diff
@@ -0,0 +1,27 @@
...
+{{% /notice %}}
+
+## Logging
+The update process will create an `update.log` file on the USB drive. If any errors occur, they will be displayed there.
+
+## Failed update
+If the update file fails, you can try again. If it fails multiple times, you may need to [flash your SD card](../sd-card).
```
PeterJCLaw: How will they know it's failed?

content/kit/pi/update-file.md
```diff
@@ -0,0 +1,27 @@
+---
+title: The update file
+---
+
+The update file is the simplest way to update your Pi with new software from us.
```
PeterJCLaw: Where do they get such a file from?

content/kit/pi/sd-card.md
```diff
@@ -0,0 +1,26 @@
+---
+title: Flashing your SD card
+---
+
+To fully update your Pi's software, or refresh it if you think it's not working correctly, you can flash our SD card image onto the microSD card in your Pi. 
+
+To update the SD card, you'll need to download our update file, `pi-image.img.xz`. The flashing procedure is identical to flashing other images.
```
PeterJCLaw: Calling `pi-image.img.xz` an update file is confusing. Can we call it something else?

PeterJCLaw: for clarity I suggest: "other images" -> "other Raspberry Pi images".

content/kit/pi/sd-card.md
```diff
@@ -0,0 +1,26 @@
...
+
+## Etcher
+We recommend using [etcher](https://etcher.io), as it's simple to use, and available on Windows, macOS and Linux. If you have another preferred tool, that will also work.
```
PeterJCLaw: Maybe change the second sentence to start: "If you're familiar with Raspberry Pis and have flashed images before with a different tool..."

content/kit/pi/update-file.md
kierdavis: In PR description:

>This PR doesnt contain the page to download files, that'll come later

Although I'm thinking it'd be better to create a download page now, even if it just says "no updates available" for the time being.

content/kit/pi/update-file.md
RealOrangeOne: The Pi reboots, as said above.

content/kit/pi/update-file.md
RealOrangeOne: Creating this page is next on my list, however will be done outside this branch

content/kit/pi/sd-card.md
```diff
@@ -0,0 +1,26 @@
...
+
+To update the SD card, you'll need to download our image, `pi-image.img.xz`. The flashing procedure is identical to flashing other Raspberry Pi images.
+
+## Etcher
+We recommend using [etcher](https://etcher.io), as it's simple to use, and available on Windows, macOS and Linux. If you're familiar with Raspberry Pi's and have flashed images before with a different tool, that will also work.
+ 
+![Etcher example](https://etcher.io/static/screenshot.gif)
+
+{{% notice note %}}
+If you choose to use a tool other than Etcher, you may need to extract the `pi-image.img.xz` to `pi-image.img`. There are many tools available for this.
```
kierdavis: IMO it would be nice to give an example here, such as [7-zip](http://www.7-zip.org/).

content/kit/pi/update-file.md
PeterJCLaw: I'd rather we didn't merge this until it's clear to readers where the updates will come from, even that's only a placeholder for the moment.

RealOrangeOne: This is handled in #97

content/kit/pi/update-file.md
PeterJCLaw: This feels rather disjoint. Could we add a step 4. to the list which is "wait for the pi to reboot"? That way it's a clear part of the main instructions that waiting is involved and that they need to be watching the pi during the update.

We could even go as far as suggesting that they include a `main.py` which contains:
``` python
import time
from robot import Robot
r = Robot()
while True:
    r.power_board.buzz(1, note='c')
    time.sleep(1.5)
```

content/kit/pi/update-file.md
PeterJCLaw: #97 looks good; can we link to those pages in this PR?

content/kit/pi/update-file.md
RealOrangeOne: I was thinking of something like that. I would actually rather bake the beeping into `runusb` in that case, so they don't have to include anything on their end besides the single file.

content/kit/pi/update-file.md
```diff
@@ -0,0 +1,27 @@
...
+
+The update file is the simplest way to update your Pi with new software from us.  
+
+{{% notice tip %}}
+Using this file is the recommended way to update your kit.
+{{% /notice %}}
+
+## Using the update file
+
+1. Put the `update.tar.xz` file on a USB drive
```
PeterJCLaw: We should clarify that they need to have at least the power-board plugged in. I'd actually like to see us remove this requirement as it complicates the process of updating pis in development, but it's a requirement for now.

content/kit/pi/update-file.md
PeterJCLaw: Either way we end up needing a power board being attached (see also above), which I'm not sure is something we really want to require.

content/kit/pi/update-file.md
RealOrangeOne: Agreed, that's an annoying _feature_. I'll add that a minimal robot setup is required

PR #97, opened 2018-02-04
Pi updates

RealOrangeOne:
A page will be created with this months release on on Tuesday, once the changelog / `pi-image` build as been finalised.

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -36,3 +36,5 @@ outputs:
   home:
     - html
     - json
+
+buildFuture: true
--- a/content/updates/_index.md
+++ b/content/updates/_index.md
@@ -1,0 +1,5 @@
+---
+title: Updates
+---
+
+Updates page content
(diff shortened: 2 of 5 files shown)


PeterJCLaw:
@RealOrangeOne why does this PR include a change to the theme?

RealOrangeOne:
@PeterJCLaw I incremented to test something, but apparently committed it accidentally. Have reverted

config.yml
```diff
@@ -36,3 +36,5 @@ outputs:
   home:
     - html
     - json
+
+buildFuture: true
```
PeterJCLaw: Not sure why this is included in this PR, though I think it's fairly harmless at the moment.

RealOrangeOne: Its so we can set the real date of an event in the future, and it be rendered in testing / preview

PR #91, opened 2018-02-02
Spherical coordinates

PeterJCLaw:
This adds documentation for the new spherical co-ordinate space implemented in https://github.com/sourcebots/robot-api/pull/61 and https://github.com/sourcebots/sb-vision/pull/23, including a diagram of the two spaces we document.

PR #101, opened 2018-02-06
Release jan 2018

RealOrangeOne:
Contains #87 and #91 

Tests fail due to missing `pi-image` links. These should be added before merge


content/updates/2018-jan.md
```diff
@@ -0,0 +1,21 @@
+---
+title: January 2018
```
PeterJCLaw: I think it's ambitious to call this the January 2018 update when we're now well in to February.

content/updates/2018-jan.md
```diff
@@ -0,0 +1,21 @@
...
+update: update  # pending pi-image
+---
+
+- Support for [custom Arduino commands](/api/servo-assembly/arduino-commands/)
+- Implement [Cartesian coordinate system](/api/vision/coordinates/#cartesian-coordinates)
+- Fix issue with camera after code terminates
+- Remove dates from logs (They were wrong anyway)
+- Remove unnecessary noisy log messages
+- Remove unnecessary `size` attribute from a marker
+- Deprecate broken `polar` coordinate system in favour of [`spherical`](/api/vision/coordinates/#spherical-coordinates)
```
PeterJCLaw: Minor: Unfortunately the link under the code highlighting of `spherical` is non-obvious. Not sure we can do anything to fix that here though.

content/updates/2018-jan.md
```diff
@@ -0,0 +1,21 @@
...
+- Flash the start LED once robot is ready (Also reduced chance of press not being registered)
```
PeterJCLaw: I think this "(Also ...)" is worth having as a separate point. While it was in the same bit of code, it's not really related from a user perspective. (Doing that also solves the grammar quandary it currently has).

Could we also clarify that it's the "start button press" that we're talking about?

content/updates/2018-jan.md
```diff
@@ -0,0 +1,21 @@
...
+- Make camera return items when indexing rather than interacting
```
PeterJCLaw: I'm not sure what this means.

Given the code changes which happened, I think you're referring to the change to `ResultList` so that doing `markers[0]` actually worked? If so, maybe we should word this as "Make indexing into a list of markers actually return a marker instance, instead of `None`"?

PeterJCLaw:
@RealOrangeOne did you or @Adimote get a chance to make the changes to link from the docs of the added features to note that they were introduced by the "Jan 2018" update? I think we discussed it, but didn't write it down anywhere.

Separately, it's occurred to me that a better process for the two other PRs which this one contained (but which ended up being unexpectedly merged into `master` directly) might have been to target them against this branch and then merge them, leaving only this one open.

RealOrangeOne:
I don't recall that discussion, but it does sound like a useful thing to note, and continue in future releases.

Yea, that branching strategy does make more sense. Oh well!

PR #103, opened 2018-02-11
Update completeness

RealOrangeOne:
- ~Add a notice about downgrading~
- State that update files are complete

### DIFF
--- a/content/updates/_index.md
+++ b/content/updates/_index.md
@@ -18,7 +18,3 @@ Once you have downloaded the file you need, refer to the documentation on
 [updating your pi](/kit/pi/#updating-your-pi) to apply the update.
 
 Each update file is a complete upgrade. Each file contains the changes of those before it. If you need to jump up multiple versions, you can do so by using the latest file.
-
-{{% notice warning %}}
-Downgrading using the update file is not supported, and may lead to odd behaviour. If for whatever reason you need to downgrade, flash the SD card.
-{{% /notice %}}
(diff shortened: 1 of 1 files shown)


content/updates/_index.md
```diff
@@ -16,3 +16,9 @@ to completely wipe your Pi for some reason.
 
 Once you have downloaded the file you need, refer to the documentation on
 [updating your pi](/kit/pi/#updating-your-pi) to apply the update.
+
+Each update file is a complete upgrade. Each file contains the changes of those before it. If you need to jump up multiple versions, you can do so by using the latest file.
+
+{{% notice warning %}}
+Downgrading using the update file is not supported, and may lead to odd behaviour. If for whatever reason you need to downgrade, flash the SD card.
```
PeterJCLaw: Is this because we're assuming that their datestamps are being used? If so, I'm pretty sure that that doesn't happen. I'm fairly sure that I've successfully used an old deb to downgrade.

RealOrangeOne: My main reason for including this was the through that an update file may only contain debs for changed packages, and so could potentially leave packages out of sync leading to dependency issues. I do think now however that's not the case. I'll remove it for now, we can add it later if the issue ever comes up

PeterJCLaw: Ah, I see. I think that that's a valid concern -- we're definitely going to be avoiding issuing updates to `runusb` at least if we can, so I imagine we'll probably skip other packages to.

Given that we're not actually shipping update files yet, I'm not particularly worried, though it would be good to record somewhere that we should either reinstate this or ensure that update files work for downgrades too. (I suspect we won't want the bother of checking, so will restore this!)

If we do reinstate this, let's leave a comment in the code explaining why so that we don't go around this loop again in the future!

RealOrangeOne:
Added to Trello: https://trello.com/c/a4k66zFD/79-check-whether-downgrading-with-update-file-is-possible

PR #104, opened 2018-02-11
Add update description for kickstart image

RealOrangeOne:
Add an update entry for the kickstart kit.

Also adapted the update templates, so `image` and `update` are optional frontmatter, and won't break if either doesn't exist. This is because we didn't ship an update file at kickstart, as the mechanism wasn't in-place yet.

### DIFF
--- a/content/updates/kickstart-2018.md
+++ b/content/updates/kickstart-2018.md
@@ -1,0 +1,8 @@
+---
+title: Kickstart 2018
+date: 2018-12-16
+hidden: true
+image: https://203-97323138-gh.circle-artifacts.com/0/pi-image.img.xz
+---
+
+This is the initial version of the SourceBots kit software, which was shipped at the kickstart for the 2018 competition.
--- a/layouts/updates/list.html
+++ b/layouts/updates/list.html
@@ -16,14 +16,14 @@
       <td>{{ .Date.Format "2006-01-02" }}</td>
       <td>
         {{ if .Params.update }}
-        <a href='{{ .Params.update }}'>Download <code>update.tar.xz</code></a>
+          <a href='{{ .Params.update }}'>Download <code>update.tar.xz</code></a>
         {{ else }}
           No update file available
         {{ end }}
       </td>
       <td>
         {{ if .Params.image }}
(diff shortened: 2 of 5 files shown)


layouts/updates/list.html
```diff
@@ -14,8 +14,20 @@
     <tr>
       <td>{{ .Title }}</td>
       <td>{{ .Date.Format "2006-01-02" }}</td>
-      <td><a href='{{ .Params.update }}'>Download <code>update.tar.xz</code></a></td>
-      <td><a href='{{ .Params.image }}'>Download <code>pi-image.img.xz</code></a></td>
+      <td>
+        {{ if .Params.update }}
+        <a href='{{ .Params.update }}'>Download <code>update.tar.xz</code></a>
```
PeterJCLaw: Could we either indent this line one more so that it matches the "No update file available" text, or dedent that line to match this one? (The same comment also applies to the image file below)

I'm not sure what our usual guidance is for intending inside template blocks; do we have a preference?

RealOrangeOne: My preference is to consider template tags like this to create a new block, and so should be indented. Will update

content/updates/kickstart-2018.md
```diff
@@ -0,0 +1,8 @@
+---
+title: Kickstart 2018
```
PeterJCLaw: While I can see why this name was chosen, it looks a bit odd when combined with the name of the January update ("January 2018"):

image

Unfortunately I can't immediately think of a name which is obviously better, though some possible names are "December 2017"; (just) "Kickstart"; "2018 Kickstart"; "Kickstart SB2018"; "SB2018 Kickstart".

content/updates/kickstart-2018.md
```diff
@@ -0,0 +1,8 @@
...
+date: 2018-12-16
```
PeterJCLaw: Whoops, just spotted that this should be "2017-12-16" (note year change)

content/updates/kickstart-2018.md
PeterJCLaw: The more I look at this the more I'm leaning towards "2018 Kickstart".

PR #106, opened 2018-02-16
Mention when features were added

PeterJCLaw:
This adds `note` boxes which explain that some of the documented features need a recent update in order to be used, along with a [shortcode](https://gohugo.io/content-management/shortcodes/) which defines the boxes.

These boxes ended up being quite a bit larger than I'd originally imagined, but I think are in-keeping with these docs (I'd originally envisioned these being a simple "New in <version>" line, like what you see in the Python docs).

Reviewers: please consider whether the boxes are suitable, both in terms of consistency and whether they're likely to be noticed.

Depends on #107.

### DIFF
--- a/layouts/shortcodes/added_in.html
+++ b/layouts/shortcodes/added_in.html
@@ -15,7 +15,7 @@
   <div class="notices warning">
     <p>
       <strong>ERROR</strong>:
-      {{ errorf "Invaild update specified: '%s' does not exist." $update }}
+      {{ errorf "Invalid update specified: '%s' does not exist." $update }}
     </p>
   </div>
 {{ end }}
--- a/layouts/shortcodes/added_in.html
+++ b/layouts/shortcodes/added_in.html
@@ -1,0 +1,21 @@
+{{ $feature := .Get "feature" }}
+{{ $update := .Get "update" }}
+
+{{ with $.Site.GetPage "page" "updates" $update }}
+  <div class="notices note">
+    <p>
+      Support for <em>{{ $feature }}</em> was added in the
+      <a href="{{ .RelPermalink }}" title="{{ .LinkTitle }}">{{ .Title }}</a>
+      update. You need to be running at least that version of the kit software
+      in order to use this feature.
+    </p>
+  </div>
+{{ else }}
+  <!-- If you see this block, you've supplied an invalid update name. -->
+  <div class="notices warning">
+    <p>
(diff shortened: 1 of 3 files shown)


PeterJCLaw:
I'm currently using:
```
$ hugo version
Hugo Static Site Generator v0.33 linux/amd64 BuildDate:
```
This project claims to depend on `>=0.30` but then actually installs `0.31.1` on the CI and on netlify.

Hugo's docs don't appear to have any notion of when features were introduced, which doesn't really help here.

@RealOrangeOne Is it possible to update to at least `0.33`?

PeterJCLaw:
Hrm, `0.32` looks like it has quite a lot of changes, that said the docs seem to have been working for me on `0.33`.

layouts/shortcodes/added_in.html
```diff
@@ -0,0 +1,21 @@
...
+      update. You need to be running at least that version of the kit software
+      in order to use this feature.
+    </p>
+  </div>
+{{ else }}
+  <!-- If you see this block, you've supplied an invalid update name. -->
+  <div class="notices warning">
+    <p>
+      <strong>ERROR</strong>:
+      {{ errorf "Invaild update specified: '%s' does not exist." $update }}
```
kierdavis: Invaild -> Invalid

layouts/shortcodes/added_in.html
```diff
@@ -0,0 +1,21 @@
+{{ $feature := .Get "feature" }}
+{{ $update := .Get "update" }}
+
+{{ with $.Site.GetPage "page" "updates" $update }}
+  <div class="notices note">
```
RealOrangeOne: It looks like there might be a way to reuse the shortcode here instead of defining the HTML manually, but it looks like some research into that is required, so for now this will do

layouts/shortcodes/added_in.html
PeterJCLaw: Thanks! https://github.com/sourcebots/docs/pull/106/commits/26646d0088155f794798d454086ab556b99c3ae5

layouts/shortcodes/added_in.html
PeterJCLaw: Yeah, I'd hoped to do so, but couldn't make it work.

PeterJCLaw:
Yeah, I wondered about a custom box type, but I think this is ok for now. If we end up with lots of these then maybe we should revisit it.

PR #107, opened 2018-02-17
Update hugo to 0.33

PeterJCLaw:
This adds support for a use of `.Get` which we want, though isn't actually the latest at the time of updating (only so that this upgrade is minimal).


RealOrangeOne:
I plan very soon to go through this repo and upgrade it to the latest versions of `hugo`, mostly to take advantage of the new features (most specifically Resources!)

PR #109, opened 2018-02-20
Migrate some content to page resources

### LINKED ISSUE #108: Directory name /static/docs is confusing
This directory name is not very clear as to what it contains.

I wouldn't have said that a pdf of markers is a document.

Perhaps /static/downloads/ or /static/pdf/

RealOrangeOne:
- Upgrade Hugo to 0.36
- Switch single-page static assets to page resources
- Add starts of image handling
- Fix #108, by deleting it
- Remove duplicate marker PDF section
- Update the theme

I don't want to move all the images over to bundles yet, as we'd end up with duplicate content, as some images are used on multiple pages. It's not _yet_ possible to access `static/` as resources, and headless bundles aren't very nice IMO. In later releases, i'll swap the rest of the content over to resources, with some helpful shortcodes obviously :tada:


trickeydan:
Have we considered using Hugo's [shortcodes](https://gohugo.io/content-management/shortcodes/) with the figure element instead of resource_img?

kierdavis:
What's the reason for upgrading Hugo? I'm assuming it's something along the lines of us wanting to use a feature or bugfix that was introduced in this version.

content/rules/_index.md
```diff
@@ -5,4 +5,4 @@ weight: 1
 
 You can find the rules for SourceBots 2018 below:
 
-# [Download the rules!](/rules.pdf)
+# {{% resource_link src="rules.pdf" %}}Download the rules{{% /resource_link %}}
```
kierdavis: :+1:

content/markers/_index.md
```diff
@@ -7,10 +7,7 @@ Our vision system uses the [AprilTag](https://april.eecs.umich.edu/software/apri
...
-| PDF                               | Marker Size | Paper Size |
-|-----------------------------------|-------------|------------|
-| [Large markers][large-marker-pdf] | 250x250mm   | A3         |
-| [Small markers][small-marker-pdf] | 100x100mm   | A4         |
-
-[large-marker-pdf]: /docs/large-tags.pdf
-[small-marker-pdf]: /docs/small-tags.pdf
+| PDF                             | Marker Size | Paper Size |
+|---------------------------------|-------------|------------|
+| [Large markers](large-tags.pdf) | 250x250mm   | A3         |
```
kierdavis: :-1: This link, and the one below, both give a 404 on the [Netlify preview](https://deploy-preview-109--sourcebots-docs.netlify.com/markers).

content/rules/_index.md
kierdavis: :+1:

content/api/vision/coordinates/index.md
```diff
@@ -52,4 +52,5 @@ robot, exactly 1 metre away.
 The following diagram shows the orientation of the Cartesian axes relative to
 the camera as well as the angles which describe the spherical coordinate space.
 
-![A diagram showing the coordinate spaces](/img/api/coordinate-spaces.svg)
+
+{{% resource_img src="coordinate-spaces.svg" alt="A diagram showing the coordinate spaces" %}}
```
kierdavis: :+1:

content/kit/motor-board/index.md
```diff
@@ -34,7 +34,7 @@ The case measures 70x84x20mm. Don’t forget that the cables will stick out.
 ## Designs
 You can access the schematics and source code of the firmware on the motor board in the following places. You do not need this information to use the board but it may be of interest to some people.
 
-- [Full Schematics](/docs/motor-schematic.pdf)
+- {{% resource_link src="motor-schematic.pdf" %}}Full Schematic{{% /resource_link %}}
```
kierdavis: :+1:

RealOrangeOne:
The main feature is page bundles, which means we can put static assets with their content to better group things. It also supports image resizing / cropping etc, but that requires moving images into page bundles, which would create duplicate content. 

There's an open issue tracking this feature, and some preliminary work https://github.com/gohugoio/hugo/issues/4381.

Obviously there's other things like additional performance, bug fixes, things like that. And of course, who doesn't like new shiny things!

content/kit/power-board/index.md
```diff
@@ -63,6 +63,6 @@ The case measures 83x99x24mm. Don’t forget that the cables will stick out.
 ## Designs
 You can access the schematics and source code of the firmware for the power board in the following places. You do not need this information to use the board but it may be of interest to some people.
 
-- [Full Schematics](/docs/power-schematic.pdf)
+- {{% resource_link src="power-schematic.pdf" %}}Full Schematic{{% /resource_link %}}
```
kierdavis: :+1:

kierdavis:
Huh, after more investigation this seems to actually be another case of links behaving differently based on whether there's a slash at the end of the URL or not.

The links at the page work fine: https://deploy-preview-109--sourcebots-docs.netlify.com/markers/
But links on this one give 404s: https://deploy-preview-109--sourcebots-docs.netlify.com/markers

content/api/vision/_index.md
```diff
@@ -25,9 +25,3 @@ The markers in the list have some useful attributes:
 {{% notice tip %}}
 A marker's position can be represented using both the [Cartesian](coordinates/#cartesian-coordinates) and [spherical](coordinates/#spherical-coordinates) coordinate systems.
 {{% /notice %}}
-
-## Marker PDFs
```
PeterJCLaw: Why are we removing this? This is a potentially breaking change for people used to coming to this page for these links.

scripts/build.sh
```diff
@@ -8,4 +8,4 @@ rm -rf public/
 rm -rf content/tutorials/kit-assembly.files
 cp -r static/img/assembly content/tutorials/kit-assembly.files
 
-hugo -v
+hugo -v --gc
```
PeterJCLaw: Curiosity: what does `--gc` do?

RealOrangeOne: It removes the build cache of static files before building the site. Resizing images is fairly trivial, and this removes any issues which may come from changes in the resizing

content/api/vision/_index.md
RealOrangeOne: It's removed mostly to remove the duplicated content. Perhaps adding a link would've been better? The markers are already a top-level section (theyre in the navbar), so finding the links isn't difficult

RealOrangeOne:
@PeterJCLaw makes some good points on file duplication in this. I didn't realise when I started how much there would be. I think it's worth revisiting this when the required updates to Hugo are made to support this, and we can properly use resources from a single location (in the cases it's needed), and not fix content to only being accessible from a single page.

(This is _sort of_ already implemented in [Headless Page Bundles](https://github.com/gohugoio/hugo/releases/tag/v0.35), but i'm not a fan of that mechanism for solving this, so it's probably better to wait for a better solution.

PeterJCLaw:
@RealOrangeOne is there anything holding us back from upgading hugo anyway? I realise that we might not (currently) need any of the latest features, but if we've already validated that it works then there might be benefit from doing so.

Related question: what's the support lifecycle of a hugo release? I realise they're still very much in the move-fast-and-break-things stage (0.x), but are they trying to keep it easy to upgrade between adjacent releases? releases, say, 5 apart?

RealOrangeOne:
@PeterJCLaw hmm, perhaps i'll strip out the resource changes from this PR and ship. 

The releases currently _may_ contain breaking changes, but all the time i've been using it, there's only been a handful, if that. As it's currently still _technically_ in beta, that's to be expected. I think it's fine to be upgrading and fixing things as and when new versions come out (not every version, they happen incredibly often), but as you say every 5 releases or so perhaps

PeterJCLaw:
@RealOrangeOne it might be easier to put up a separate PR for the upgrade than to try to munge this one.

PeterJCLaw:
I'm closing this PR in favour of #147 (which declares that it replaces it).

PR #110, opened 2018-02-20
Fix missing peer dependency that was preventing building.

trickeydan:
Also add package-lock.json to the .gitignore file

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -129,6 +129,8 @@ typings/
 # dotenv environment variables file
 .env
 
+# NPM Package Lock File
+package-lock.json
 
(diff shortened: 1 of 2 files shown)


RealOrangeOne:
I think the issue here is more that `sinon` is a peer dependency, rather than us needing to rely on it. Peer dependencies also don't prevent installing, so I'm not exactly sure why it's blowing up (it'll throw an error, but still build perfectly fine)

Perhaps this fix is better made upstream?

trickeydan:
I've just found that I had the wrong version of Hugo installed.

PR #111, opened 2018-02-20
Upgrade to Node 8

RealOrangeOne:
Upgrade to node 8, so we can use the `package-lock.json` feature to prevent unpinned dependencies breaking our builds, as they do.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ version: 2.0
 jobs:
     build:
         docker:
-            - image: circleci/node:6.10
+            - image: circleci/node:8.9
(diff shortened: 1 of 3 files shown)


kierdavis:
Why does the package.lock need to be stolen from the CI? My limited knowledge of npm suggests that it should be created when running `npm install` locally for the first time, but I could be wrong.

RealOrangeOne:
The issue is that a dependency of `markdown-spellcheck` wasn't pinned properly, and so was a different version when installed on newer systems, which broke the tests. The CI has a cache which meant it didnt break, so I stole the lock file from there to use here. 

Some day i'll actually look into what went wrong, but that day isn't today!

PR #117, opened 2018-02-27
Add links to kit items.

trickeydan:
Add links to kit items. Have attempted to get as close to the kit we provide if I was unable to find the exact item.

### DIFF
--- a/content/kit/_index.md
+++ b/content/kit/_index.md
@@ -26,20 +26,32 @@ With your kit, you should have the following boards:
 
 In addition to the boards, your kit will also contain:
 
-| Part                                                                                                     | Quantity | Specification                 | Notes                                            |
-|----------------------------------------------------------------------------------------------------------|----------|-------------------------------|--------------------------------------------------|
-| [Battery Bag](https://hobbyking.com/en_us/lithium-polymer-charge-pack-25x33cm-jumbo-sack.html)           | 1        |                               |                                                  |
-| [Battery Charger](https://hobbyking.com/en_us/imax-b6-50w-5a-charger-discharger-1-6-cells-genuine.html)  | 1        | iMAX B6                       |                                                  |
-| [Battery Charger Power Supply](https://www.amazon.co.uk/TMEZON-100V-240V-Switching-Security-Accessories-1x-5A-power-adapter/dp/B01N1ULL37) | 1      | 12V 5A                                           |
-| [Battery](https://hobbyking.com/en_us/turnigy-2200mah-3s-25c-lipo-pack.html)                             | 2        | 11.1V 3S 25C 2200mAh Lithium Polymer |                                           |
-| [Webcam](https://www.amazon.co.uk/TeckNet%C2%AE-C016-Webcam-MegaPixel-Microphone/dp/B000Q3VECE/)         | 1        | TeckNet C016                  |                                                  |
-| [USB Hub](https://www.amazon.co.uk/Warranty-Ultra-Slim-Portable-Expansion-Splitter-Black/dp/B01DYD3Q28/) | 1        |                               |                                                  |
-| [Micro USB Cable](https://www.amazon.co.uk/AmazonBasics-USB-Male-Micro-Cable/dp/B01EK87T9M/)             | 2        |                               |                                                  |
-| [USB B Cable](https://www.amazon.co.uk/AmazonBasics-USB-2-0-Cable-Male/dp/B00NH11KIK)                    | 1        |                               |                                                  |
-| [Large CamCon](http://uk.farnell.com/camdenboss/ctb9400-2/terminal-block-pluggable-2pos/dp/3882275)      | 2        | 7.5mm                         |                                                  |
-| [Medium CamCon](http://uk.farnell.com/camdenboss/ctb9200-2a/terminal-block-pluggable-2pos/dp/3881854)    | 2        | 5mm                           |                                                  |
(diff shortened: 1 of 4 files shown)


RealOrangeOne:
Why are we using googles tracked links for these? It'd be nicer to just use the full links

trickeydan:
I will switch to the full urls and hopefully that will fix the CI tests

kierdavis:
From the spam in the CI log it looks like it's timed out waiting for a response when trying to check the links.

And yeah I agree, it'd be nicer to use the true URLs rather than goo.gl shortened ones. It's not like we're asking them to type the URLs in manually or anything.

content/kit/_index.md
```diff
@@ -26,20 +26,20 @@ With your kit, you should have the following boards:
...
+| [USB B Cable](https://www.amazon.co.uk/AmazonBasics-USB-2-0-Cable-Male/dp/B00NH11KIK)                    | 1        |                               |                                                  |
+| [Large CamCon](http://uk.farnell.com/camdenboss/ctb9400-2/terminal-block-pluggable-2pos/dp/3882275)      | 2        | 7.5mm                         |                                                  |
+| [Medium CamCon](http://uk.farnell.com/camdenboss/ctb9200-2a/terminal-block-pluggable-2pos/dp/3881854)    | 2        | 5mm                           |                                                  |
+| [Small CamCon](http://uk.farnell.com/camdenboss/ctb92he-2/terminal-block-pluggable-2pos/dp/1717047)      | 3        | 3.81mm                        | 1 additionally found on Raspberry Pi power cable |
+| Raspberry Pi Power Cable                                                                                 | 1        | Micro USB to small CamCon     |                                                  |
```
antonnikitin97: There's no link here (Should there be???)

kierdavis: The Raspberry Pi power cable is a custom part we make ourselves, so there isn't really a page online describing it.

kierdavis:
Pro tip: Amazon URLs can be shortened by replacing the stuff between `amazon.co.uk/` and `/dp/` with something shorter. For example:

    https://www.amazon.co.uk/AmazonBasics-USB-2-0-Cable-Male/dp/B00NH11KIK
    https://www.amazon.co.uk/-/dp/B00NH11KIK

are both equivalent.

It's not really worth changing it unless you really want to though.

content/kit/_index.md
antonnikitin97: fair enough :)

trickeydan:
@RealOrangeOne Any suggestions for how this could be fixed?

package.json
```diff
@@ -5,7 +5,7 @@
   "scripts": {
     "spellcheck": "mdspell --en-gb --report --ignore-acronyms --ignore-numbers --no-suggestions 'content/**/*.md'",
     "broken-link-local": "blcl --recursive --ordered --exclude-external ./public/",
-    "broken-link-all": "blcl --recursive --ordered ./public/",
+    "broken-link-all": "blcl --recursive --exclude farnell --ordered ./public/",
```
PeterJCLaw: Why are we excluding farnell here?

content/kit/_index.md
```diff
@@ -26,20 +26,20 @@ With your kit, you should have the following boards:
...
-| Large CamCon                 | 2        | 7.5mm                         |                                                  |
-| Medium CamCon                | 2        | 5mm                           |                                                  |
-| Small CamCon                 | 3        | 3.81mm                        | 1 additionally found on Raspberry Pi power cable |
-| Raspberry Pi Power Cable     | 1        | Micro USB to small CamCon     |                                                  |
+| Part                                                                                                     | Quantity | Specification                 | Notes                                            |
+|----------------------------------------------------------------------------------------------------------|----------|-------------------------------|--------------------------------------------------|
+| [Battery Bag](https://hobbyking.com/en_us/lithium-polymer-charge-pack-25x33cm-jumbo-sack.html)           | 1        |                               |                                                  |
```
PeterJCLaw: Style suggestion: footnote style links _might_ make this table easier to read:

``` markdown
| [Battery Bag][battery-bag]           | 1        | ... |

[battery-bag]: https://hobbyking.com/en_us/lithium-polymer-charge-pack-25x33cm-jumbo-sack.html

package.json
trickeydan: Farnell links are excluded as they cause the link checker to hang indefinitely. See earlier build https://circleci.com/gh/sourcebots/docs/535 It is not entirely clear why this is the case. One theory was that the User Agent for the link checker was being filtered, but this is not the case.

PeterJCLaw: Please could we have an inline comment to that effect?

PeterJCLaw: Meta (and well beyond the scope of this PR): this is a class of error which I see as unavoidable and a reason not to have external link checking.

content/kit/_index.md
trickeydan: I am unable to find a reference to this syntax in the Hugo documentation. Please could you provide a link? It certainly seems like a good idea, as having very long urls in the table feels inelegant

PeterJCLaw: This style of links (search for "Reference-style links") is part of the original [markdown syntax](https://daringfireball.net/projects/markdown/syntax#link), so I don't know whether it would be documented explicitly by hugo.

package.json
trickeydan: Inline comment added: de78cf8

content/kit/_index.md
trickeydan: Changed in 79586e4

trickeydan:
As teams are asking for this now (at the tech day), I suggest this is merged and the package.json is fixed in a separate PR

PR #123, opened 2018-03-04
Improve formatting on Introduction to Python

trickeydan:
I know that when this was discussed in the meeting, use of a table was proposed. 
However, I decided that it would be best to keep the formatting consistent with the Tutorials index page.


kierdavis:
Thanks :+1:

PR #124, opened 2018-03-06
Fix missing apostrophe

trickeydan:


trickeydan:
This appears to be failing due to https://sublimetext.com/ being down.

I suggest we change our workflow in the future such that it is not reliant on external websites being down

PR #125, opened 2018-03-06
Add Github Edit Links

trickeydan:
When these were recently removed in 1add5209f53d5a09c8b264f15d7e55492be3eaac, the reason cited was that it broke the link checker. I have fixed this by excluding those links from the checker


package.json
```diff
@@ -5,7 +5,7 @@
   "scripts": {
     "spellcheck": "mdspell --en-gb --report --ignore-acronyms --ignore-numbers --no-suggestions 'content/**/*.md'",
     "broken-link-local": "blcl --recursive --ordered --exclude-external ./public/",
-    "broken-link-all": "blcl --recursive --ordered ./public/",
+    "broken-link-all": "blcl --recursive --exclude https://github.com/sourcebots/docs/tree/master/content/* --ordered ./public/",
```
PeterJCLaw: Please could we have an inline comment explaining the exclusion.

Meta: this will likely conflict with #117.

trickeydan: Please see above note about @RealOrangeOne's comments in 1add520

PeterJCLaw: I understand the reason for the exclusion; I'm asking for a comment so that it's clear to all future readers of the code.

Please bear in mind that future readers _won't_ have to hand the comments on this PR or the original commit which removed the functionality. While they _could_ `git blame`, the chances are that they'd have to chase several layers of commits to find the rationale (and then they'd need to work backwards from "Add Github Edit Links" to this PR and onwards from there). Adding a comment now is easy and makes things clearer for everyone.

trickeydan: Ah, I missed the word inline and mistook the context of your request.

Fixed in 9d037ff

package.json
```diff
@@ -5,7 +5,8 @@
   "scripts": {
     "spellcheck": "mdspell --en-gb --report --ignore-acronyms --ignore-numbers --no-suggestions 'content/**/*.md'",
     "broken-link-local": "blcl --recursive --ordered --exclude-external ./public/",
-    "broken-link-all": "blcl --recursive --ordered ./public/",
+    "//": "Github links below are excluded to prevent breaking the link checker for the Edit in Github urls, when no such page exists in master",
```
kierdavis: Eurgh, does npm's JSON parser really not support comments?

trickeydan: https://groups.google.com/forum/#!msg/nodejs/NmL7jdeuw0M/yTqI05DRQrIJ

RealOrangeOne: ah, i never saw this hack. I'd rather remove this comment

PeterJCLaw:
Hnng. Why is the JavaScript ecosystem so crap.

This is a difficult one. I really want there to be more help to future readers of the `package.json` about why certain things are ignored than the minimal commit messages we have here. However I agree the that comment hack is _really_ nasty.

Is there another way that we can configure the exclusions? Perhaps a config file for the tool? Alternatively, could we break that command out to a script (in the `scripts` directory) and have the comments there?

RealOrangeOne:
The tool has no config, although perhaps extracting it out into `scripts/test.sh` might be better, as we can have tests there. Obviously you'll need to expand path with `node_modules/.bin/`

RealOrangeOne:
wrong button -_-

PeterJCLaw:
I think pulling out a `./scripts/check-all-links` (with suitably path munging) is reasonable. If we pulled out all the testing things to a single script I'd be in favour of removing them from the `package.json` to avoid confusion.

trickeydan:
As mentioned yesterday at the tech day, the scripts are out of scope of this PR.

RealOrangeOne:
It looks like due to a merge issue, `farnel` has been removed from the exclusion list. Restoring this should fix the tests

PR #129, opened 2018-03-09
Tag the latest update

Adimote:
Add a `(Latest)` to the latest update to improve readability

### DIFF
--- a/layouts/updates/list.html
+++ b/layouts/updates/list.html
@@ -11,7 +11,7 @@
   </tr>
   {{ range $index, $_ := .Pages.ByDate.Reverse }}
     <tr>
-      <td><a href='{{ .RelPermalink }}' title="See a summary of the changes in this release">{{ .Title }}</a>{{ if (eq $index 0) }}(Latest){{ end }}</td>
+      <td><a href='{{ .RelPermalink }}' title="See a summary of the changes in this release">{{ .Title }}</a>{{ if (eq $index 0) }} (Latest){{ end }}</td>
       <td>{{ .Date.Format "2006-01-02" }}</td>
       <td>
         {{ if .Params.update }}
--- a/layouts/updates/list.html
+++ b/layouts/updates/list.html
@@ -9,9 +9,9 @@
     <th>Update File Download</th>
     <th>Image Download</th>
   </tr>
-  {{ range .Pages.ByDate.Reverse }}
+  {{ range $index, $_ := .Pages.ByDate.Reverse }}
     <tr>
-      <td><a href='{{ .RelPermalink }}' title="See a summary of the changes in this release">{{ .Title }}</a></td>
+      <td><a href='{{ .RelPermalink }}' title="See a summary of the changes in this release">{{ .Title }}</a>{{ if (eq $index 0) }}(Latest){{ end }}</td>
       <td>{{ .Date.Format "2006-01-02" }}</td>
       <td>
         {{ if .Params.update }}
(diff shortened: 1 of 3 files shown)


layouts/updates/list.html
```diff
@@ -9,9 +9,9 @@
     <th>Update File Download</th>
     <th>Image Download</th>
   </tr>
-  {{ range .Pages.ByDate.Reverse }}
+  {{ range $index, $_ := .Pages.ByDate.Reverse }}
     <tr>
-      <td><a href='{{ .RelPermalink }}' title="See a summary of the changes in this release">{{ .Title }}</a></td>
+      <td><a href='{{ .RelPermalink }}' title="See a summary of the changes in this release">{{ .Title }}</a>{{ if (eq $index 0) }}(Latest){{ end }}</td>
```
trickeydan: There is a space missing between the update name and (Latest).

trickeydan: I also think it could be nice to have the text (Latest) in a different colour to highlight it.

Adimote: Fixed in d3101fb, I think colouring it might be a bit tacky.

kierdavis:
Since the release dates are clearly visible, I don't feel this is necessary.

Adimote:
It's not necessary but it really improves usability. It's not like it will break in the near future

Adimote:
We need to think about the user here, **having some bold text pointing where people should look is very helpful**

PeterJCLaw:
> We need to think about the UX here,

This is a perfectly reasonable point. However it's non-obvious to me that this implementation is the best. Personally I think it adds visual clutter which decreases the UX.

We should also consider the technical overhead of doing this. No change that adds things is free of maintenance or cognitive load when reading the source. All features are forms of complexity. We should be judging whether that complexity is worth the benefits of the feature.

> *having some bold text pointing where people should look is very helpful.*

I agree with that, though I'd generalise it that visual markers of some form can help point things out.
How about making the link (or maybe even the whole row) bold rather than adding more text?
If we're doing the whole row, how about a slightly different background colour?

> Especially because there's no obvious order for the update name on the far left column, some may think that 'Kickstart 2018' is the latest.

I disagree. Everyone in the target audience knows what Kickstart was and _when_ it was. That the rest are named after dates means that it's pretty obvious what the ordering is.
Even if that weren't the case, then the "Release date" column is pretty clear about the ordering. I think it's reasonable to expect people to read the whole table while scanning from left to right looking for the download link.

PeterJCLaw:
I think the ideal implementation would add a class to the row of the current update and then that would be styled. This nicely separates content from styling, yet still allows for any of the visual implementations which have been proposed so far.

Adimote:
Tests are failing due to an unrelated link check

RealOrangeOne:
>  there's no obvious order for the update name on the far left column

I'm not sure if this was done before this PR was made, but the release date is in the list, so it's obvious which way round it is. The names are also somewhat-related to when they were released (I don't like this, it's on my list!). But I think adding additional noise isnt the answer.

An idea I was toying with on something external, was having a specific entry outside the table, for the latest, and further down, having a table of all releases. Much like the buttons we have at the bottom of the update page, we specifically move the latest release buttons out of the table flow, making it incredibly obvious, without adding noise to the table. Most versioned software downloads employ a similar approach.

trickeydan:
@Adimote Same broken link issue occurred in #125, have excluded microsoft links.

PR #132, opened 2018-03-17
Remove reference to update file

RealOrangeOne:
Remove reference to update file, as we haven't shipped one yet. 

Restoring should be done by reverting this PR.


content/kit/pi/_index.md
```diff
@@ -18,13 +18,6 @@ Please do not attempt to modify the software on the Raspberry Pi. It has been co
 ## Updating your Pi
 Keeping your kit up to date is very important. It enables us to deploy new features, as well as fix bugs.
 
-There are 2 ways of updating your kit:
-
-- [Using an `update.tar.xz` file](update-file)
-- [Flashing your SD card](sd-card)
+We have documentation on [Flashing your SD card](sd-card).
```
PeterJCLaw: This reads a little oddly, perhaps this?:
> You can update your Pi by [flashing your SD card](sd-card).

content/updates/_index.md
```diff
@@ -2,19 +2,9 @@
...
-The update files are the easiest way to apply an update and are the recommended
-approach. The full images are provided in case an update goes wrong or you need
-to completely wipe your Pi for some reason.
+Keeping your kit up to date is very important. It enables us to deploy new features, as well as fix bugs. The following table outlines the updates which have been published.
 
 Once you have downloaded the file you need, refer to the documentation on
 [updating your pi](/kit/pi/#updating-your-pi) to apply the update.
 
-Each update file is a complete upgrade. Each file contains the changes of those before it. If you need to jump up multiple versions, you can do so by using the latest file.
+Each file contains the changes of those before it. If you need to jump up multiple versions, you can do so by using the latest file.
```
PeterJCLaw: I think we can probably just drop this -- it was thought to be needed as the behaviour of the update files is non-obvious. The behaviour of flashing the Pi is somewhat clearer IMO.

(Not sure if I'm biased by know what the flashing is doing though?)

If we do want to keep this, then maybe explain why the statement is the case? Perhaps something like:
> Flashing an image file replaces all the software on the Pi with the versions in the image, so you only need to install the one that you want to use.

Even then, I think that reads oddly as we're omitting a detail about why that comment is needed.

Issue #118, opened 2018-03-01
Potential for confusion between digit '1' and lowercase 'l'


kierdavis:
As with many fonts, the digit '1' and the lowercase letter 'l' appear quite similar in the font we use for code samples, and can be easily confused. It may be easy for experienced developers to infer from the context which character is in use in a particular situation, but this may be difficult for those who have seen very little code before.

@RealOrangeOne suggested that the syntax highlighting would be sufficient for distinguishing the two, but I disagree for a couple of reasons:

* this doesn't account for identifiers containing both letters and numbers
* it assumes that none of our users are colourblind.

I would recommend switching to a different font in which this is less of a problem.

Spun out from #116.

PR #134, opened 2018-03-27
Replace Consolas with Courier New

trickeydan:
This PR addresses #118 as discussed in the meeting earlier today.

### DIFF
--- a/layouts/partials/custom-header.html
+++ b/layouts/partials/custom-header.html
@@ -1,0 +1,1 @@
+<link href="/css/style.css" rel="stylesheet">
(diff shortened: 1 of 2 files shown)


trickeydan:
Tests appear to be failing due to issues with the Link Checker :roll_eyes:

prophile:
What's the licensing on these fonts

RealOrangeOne:
Also, are these fonts installed on the various key OSs by default? It might be worth checking the output on a windows machine (7 and 10) to check it actually works as we want it to.

trickeydan:
I have not checked on other machines, but I have copied the font stack from [CSS Font Stack](https://www.cssfontstack.com/Courier-New).
Install base for `Courier New` is reported as  Win: 99.73%  Mac: 95.68%. As it is a stack of fonts, it will fall back to `monospace` if the font is missing anyway so should not result in a bad render.

PR #136, opened 2018-04-17
Release 2018.3.0

RealOrangeOne:
Missing image link from https://github.com/sourcebots/pi-image/pull/24

### DIFF
--- a/content/api/game-state.md
+++ b/content/api/game-state.md
@@ -1,0 +1,23 @@
+---
+title: Game State
+---
+
+You can access information on the current state of the game.
+
+## Mode
+Your robot can be in 2 modes, `DEVELOPMENT` and `COMPETITION`. During competition mode, your robot will stop executing at the end of the match.
+
+```python
+from robot.game import GameMode
+
+r.zone == GameMode.DEVELOPMENT
+>> True
+```
+
+## Zone
+The corner your robot starts in, known as its _zone_, is stored in the zone property. This is `0` by default.
+
+```python
+r.zone
+>> 0
+```
--- a/content/api/game-state.md
+++ b/content/api/game-state.md
@@ -2,22 +2,24 @@
 title: Game State
 ---
 
-You can access information on the current state of the game.
-
 ## Mode
-Your robot can be in 2 modes, `DEVELOPMENT` and `COMPETITION`. During competition mode, your robot will stop executing at the end of the match.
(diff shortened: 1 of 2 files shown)


content/api/game-state.md
```diff
@@ -0,0 +1,25 @@
...
+from robot.game import GameMode
+
+r.mode == GameMode.DEVELOPMENT
+>> True
+```
+
+During competition mode, your robot will stop executing code at the end of the match.
+
+## Zone
+Your robot will start in a corner of the arena, known as its starting zone. The number of zones depends on the game. Each zone is given a number, which you can access with the `zone` property:
```
antonnikitin97: is this only available in competition mode? if so what are the consequences of accessing this property in dev mode?

content/api/game-state.md
```diff
@@ -0,0 +1,25 @@
+---
+title: Game State
+---
+
+## Mode
+Your robot can be in 1 of 2 modes: `DEVELOPMENT` and `COMPETITION`. By default, your robot will be in `DEVELOPMENT` mode. To work out which mode you're in, you need to compare the values against the `GameMode`:
+
+```python
+from robot.game import GameMode
```
PeterJCLaw: Is this not re-exported by the `robot` module?
I think we should encourage top-level module use only.

content/updates/2018-03.md
```diff
@@ -0,0 +1,10 @@
+---
+title: April 2018
+version: 2018.3.0
+date: 2018-04-17
+hidden: true
+image: https://270-97323138-gh.circle-artifacts.com/0/pi-image.img.xz
+---
+
+- Allow setting the starting zone of a robot using a USB drive. This also sets the robot to competition mode.
```
PeterJCLaw: Maybe describe as a "competition USB drive" to prevent people trying to do this themselves and failing?

content/api/game-state.md
RealOrangeOne: ah, yes, you are correct.

content/updates/2018-03.md
RealOrangeOne: Will add. However it's undocumented, and we're already checking if a `main.py` exists on the drive, so we _should_ be fine

PeterJCLaw: I realise that's why it will fail, my concern is that that failure will confuse the competitors.

PR #137, opened 2018-04-18
update rules

Adimote:
update the rules with the amendments


Adimote:
I don't think this is needing review, as the rules themselves were already approved

RealOrangeOne:
I disagree completely. The rules document should be reviewed to make sure the right one is being added, with the updates we expect. As the commit was made before the PR for the rules was approved and merged, this document is clearly still out of date. We should also not have merged it, as it's now updated the rules on the site, without having sent the justification email.

PR #140, opened 2018-06-05
Bump Hugo version to 0.40

trickeydan:


mildlyincompetent:
Why is there a shell.nix file anyway? As far as I know we are not using Nix anywhere in production nor specifically recommending it as a dev tool. While everyone is obviously free to use whatever dev tools they want, I don't see why files from them should be in the repo and supported by us? I don't put my `.idea` directory in java/python repos, and I don't see how this is an different?

Adimote:
We discussed this in the meeting, and decided it's probably a good idea to upgrade hugo and leave the .nix file behind, as there are some new features we want and not many people use nix

kierdavis:
Why is a bump to the minimum required version of Hugo needed? Is there a particular feature introduced in 0.40 that we want to use?

Adimote:
@trickeydan said there was, though he didn't say what

trickeydan:
Please see Hugo changelog, lots of nice features added. I have had trouble before on a PR (cannot remember which one), where the version of Hugo that we were using on here was too old and it didn't pass tests

RealOrangeOne:
I tend to agree with @kierdavis here. If we don't need to upgrade, let's not. Until we start actually using or needing the features, or the site no longer builds. If we're going to upgrade, we might as well take the time to actually use the new features as well.

As a side note, this PR is now technically useless, as the latest is now `0.42.1`, making this bump pointless. I'll spend some time soon taking advantage of the new features released in these versions, so the bump isn't as useless.

trickeydan:
I bumped to `0.40` as that is version currently in the Debian Stretch repos. It also contains the features that I think we will want to use. It would be excessive to upgrade at every version. I suggest upgrading every 5 subversions e.g `0.45`, or when a major new feature / fix is included (that we want in the site).

PR #142, opened 2018-07-23
Remove portable python link

trickeydan:
Portable Python is no longer developed.

### DIFF
--- a/content/tutorials/python-whirlwind-tour.md
+++ b/content/tutorials/python-whirlwind-tour.md
@@ -23,7 +23,7 @@ print("Hello World!")
 
 The text `Hello World!` should appear in the output box.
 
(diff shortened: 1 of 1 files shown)


mildlyincompetent:
I agree with removing portable python, but I do think keeping an alternative to installing python is wise. Maybe link pythonanywhere or a similar provider instead, as opposed to simply removing the comment?

trickeydan:
To clarify, I was unable to find an implementation of portable python. If somebody is able to find a link, we should include it.

RealOrangeOne:
The portable python site itself suggests some more. The closest of which is https://winpython.github.io/. I don't have access to a windows machine to see if it's a good fit though

PR #143, opened 2018-07-23
Improve explanation of `in`

mildlyincompetent:
`in` works on things that aren't just strings.

### DIFF
--- a/content/tutorials/python-whirlwind-tour.md
+++ b/content/tutorials/python-whirlwind-tour.md
@@ -647,11 +647,16 @@ These return a boolean (`True` or `False`) value, and are used in `if` statement
 :   less than, less than or equal to, greater than, etc.
 
 `in`
-:   returns true if the item on the left is contained in the item on the right. The items can be strings, arrays, or other objects. For example:
+:   returns true if the item on the left is contained in the item on the right. The items can be strings, lists, or other objects. For example:
 
 ```python
 if "car" in "Scarzy's hair":
     print("Of course.")
+```
+
+```python
+if 7 in [2, 35, 7, 8]:
+    print("Found a seven!")
 ```
 
 #### Logical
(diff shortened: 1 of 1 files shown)


content/tutorials/python-whirlwind-tour.md
```diff
@@ -647,7 +647,7 @@ These return a boolean (`True` or `False`) value, and are used in `if` statement
 :   less than, less than or equal to, greater than, etc.
 
 `in`
-:   returns true if the string on the left is contained in the string on the right. For example:
+:   returns true if the item on the left is contained in the item on the right. The items can be strings, arrays, or other objects. For example:
```
RealOrangeOne: It might be worth explaining what 'contained in' means in this context. Alternatively expanding the example to contain lists and dictionaries?

nmalky: It is defined in https://docs.python.org/dev/reference/expressions.html#membership-test-details.

To summarise (paraphrasing linked docs):
- `in` and `not in` test for membership.
- `x in s` evaluates to `True` if `x` is a member of `s`, and `False` otherwise.
- For strings, `x in y` is `True` if and only if `x` is a substring of `y`.
- For user-defined classes which define the __contains__() method, x in y returns True if y.__contains__(x) returns a true value, and False otherwise.
- Otherwise, `__iter__()` is used, and finally `__getitem__()`

nmalky: Probably worth expanding to give examples for each data type (string, list, set, dict) for this tutorial.

PeterJCLaw: I don't think we need to include that level of detail -- this is a _whirlwind tour_ of Python and not meant to be exhaustive docs. (Also compare to the other explanations in the tour)

Adimote:
(merged with permission of @mildlyincompetent )

PR #145, opened 2018-07-29
Remove all updates

PeterJCLaw:
As the SourceBots competition is over, these don't mean anything any longer.


RealOrangeOne:
I would argue keeping the changelog would be useful, it's good to see the progression of the kit. I do think the links inside the documentation pages themselves can go.

PR #146, opened 2018-07-31
Update docs to match new Arduino Firmware

trickeydan:


content/api/servo-assembly/arduino-commands.md
```diff
@@ -106,42 +106,38 @@ the command.
...
 static const CommandHandler commands[] = {
   // ... lots of existing commands ... (make sure you leave them in place!)
-  CommandHandler("echo", &echo, "echoes its arguments, each on a new line"),
+  CommandHandler('E', &echo),
 };
 ```
 
-The first argument is a quoted `String` literal containing the name you'll use
-in Python to call this command. It can contain letters, numbers and dashes (`-`).
+The first argument is a char containing the name you'll use
```
RealOrangeOne: Can we either change this to character, or wrap it in backticks to show it's technical

PR #163, opened 2021-08-09
Replaced table syntax

tpoomlmly:


jefftheprogrammer:
I didn't realise you already worked on it haha

I don't think you need to make any changed anymore.

PR #166, opened 2021-08-09
Reorder Pages in index

### LINKED ISSUE #159: Information architecture gripe
Fancy phrases make me seem more intelligent

I don't think the index on the left shows the pages in the best order. Clicking the right arrow on the homepage leads to the API intro and then the Arduino API, which isn't intuitive to me. I think the Tutorials should be at the top, followed by the API, then Rules -> FAQs -> Kit.

The basic movement tutorial is good to have at the top of the tutorial page, but since we're already giving them a Python booklet I think the whirlwind tour should be at the bottom and the simulator guide should be in the middle. I expect most of the students will want to carry on programming past the 4pm finish every day.

Also, I think the API page should go Motor -> Servo -> Arduino -> Power -> Game State.

---

The order of a page can be changed with the weight field at the top of the page:
```markdown
---
title: FAQs
weight: 10
---
```
Heavier weights sink to the bottom of the index, no weight at all is the heaviest, and equal weights are alphabetically sorted.

tpoomlmly:
Closes #159

### DIFF
--- a/content/api/_index.md
+++ b/content/api/_index.md
@@ -1,6 +1,6 @@
 ---
 title: API
-weight: 1
+weight: 3
 ---
 
 Programming your robot is done in [Python](https://www.python.org/),
--- a/content/api/arduino.md
+++ b/content/api/arduino.md
@@ -1,5 +1,6 @@
 ---
 title: Arduino API
+weight: 5
 ---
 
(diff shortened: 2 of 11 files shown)


jefftheprogrammer (approved):
Looks good to me.

PR #173, opened 2021-08-13
Add virtual Arduino docs

tpoomlmly:
Covered simulator differences, analogue + digital IO.

Need to make changes to `arduino.md:31` to match @Adimote 's changes once the virtual robot docs are complete

### DIFF
--- a/.spelling
+++ b/.spelling
@@ -70,3 +70,7 @@ vcc
 VSCode
 W3Schools
 Webots
+www.python.org
+python-396
+_f_
+×
--- a/content/api/arduino.md
+++ b/content/api/arduino.md
@@ -75,8 +75,8 @@ Distance sensors are connected to 1 analogue pin each and measure the
 distance to the object ahead of them in metres. You read them like any
 other analogue pin.
 
-The value returned will be between 0 and 2 - anything further than 2m
-will be capped at 2.
+The value returned will be between 0 and 2 - anything further than 2
+metres will be capped at 2.
 
 ## Pin Mode (Unavailable in Simulator)
 
--- a/content/api/arduino.md
+++ b/content/api/arduino.md
@@ -21,7 +21,19 @@ from microswitches to LEDs. GPIO is only available on pins 2 to 13 and
 A0 to A5 because pins 0 and 1 are reserved for communication with the
 rest of our kit.
 
-## Pin mode
+## Simulator
+
+In the simulator, the Arduino's pins are pre-populated and pre-configured.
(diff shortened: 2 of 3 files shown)


content/api/arduino.md
```diff
@@ -21,7 +21,64 @@ from microswitches to LEDs. GPIO is only available on pins 2 to 13 and
...
 
-## Pin mode
+## Simulator
+
+In the simulator, the Arduino's pins are pre-populated and pre-configured.
+The first few digital pins are occupied by digital inputs, the next few by
+digital outputs, and the analogue pins are attached to distance sensors.
+
+To find out how many inputs and outputs each type of robot has, check the
+[robot docs](TODO_change_when_this_page_is_added).
```
Adimote, line 31: added in #174  :)

content/api/arduino.md
```diff
@@ -113,7 +170,7 @@ The values are the voltages read on the pins, between 0 and 5.
 Pins `A4` and `A5` are reserved and cannot be used.
 {{% /notice %}}
 
-## Ultrasound Sensors
+## Ultrasound Sensors (Unavailable in Simulator)
```
Adimote, line 173: I think we're calling the distance sensors ultrasound sensors in the simulator (they have an ultrasound sensor model, and behave like an ultrasound sensor, except they don't have a minimum distance).

Adimote (approved):

content/api/arduino.md
RealOrangeOne: Correct. The API is also slightly different.

.spelling
```diff
@@ -61,4 +61,7 @@ vcc
 VSCode
 W3Schools
 Webots
+www.python.org
```
RealOrangeOne, line 64: These items don't appear to be used in this PR?

content/api/arduino.md
```diff
@@ -21,7 +21,64 @@ from microswitches to LEDs. GPIO is only available on pins 2 to 13 and
...
+Arduino's analogue input pins starting from pin `A0`. You can read their
+values like this:
+
+```python
+distance = r.arduino.pins[AnaloguePin.A0].analogue_value
+```
+
+The value read is returned as a float.
+
+### Distance Sensor
```
RealOrangeOne, line 72: I think we need to document here how the physical robot API works for this, as it's slightly different.

.spelling
tpoomlmly: they're not, but my local copy of the spellchecker was flagging them up

RealOrangeOne (approved):

PR #181, opened 2021-08-16
Add new rules

prophile:
As discussed in the `smallpeice-rules` sim.


RealOrangeOne (approved):
I am happy enough to get this out there

PR #187, opened 2021-08-17
Add missing trailing slash to this robots docs link

PeterJCLaw:
Without the trailing slash, links on the robots page don't work properly. We should definitely fix that, however this works around the issue for now.


RealOrangeOne (commented):
:shipit:

PR #192, opened 2021-08-19
Fix radius

### LINKED ISSUE #191: Wrong wheel radius
The docs say the wheel radius is 15cm, but it's the motors that are 15 _mm_

the wheels have a 5cm radius

tpoomlmly:
Fixes #191

### DIFF
--- a/.spelling
+++ b/.spelling
@@ -16,6 +16,7 @@
 3.81mm
 3m
 4mm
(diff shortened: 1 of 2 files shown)


jefftheprogrammer (dismissed):
I looked at the proto file and I approve

PR #203, opened 2023-06-24
Move documentation to `mkdocs`

RealOrangeOne:
We currently use Hugo for our documentation, but the theme we use is no longer supported. 

Instead, I've ported to `mkdocs`, which arguably looks nicer, is incredibly well supported, and is all Python, which is easier for us to work with.

This PR adds the basics, does a little cleanup, and removes everything Hugo. There are content changes which need making based on the content here, but that's out of scope for this PR. I've removed entire defunct sections, but not gone much further than that, so we can review the content as a whole in time.

Note this PR merges into `sp2023`, which will serve as a feature branch for this year's Smallpeice.


WillB97 (approved):
Theme looks fine, lets get this merged so we can start updating the content.

PR #218, opened 2023-08-06
The rules are from 2023

tpoomlmly:
Updated "You can find the rules for Smallpeice 2021 competition here!" to 2023


RealOrangeOne:
There's an existing PR https://github.com/sourcebots/docs/pull/215, which covers this change and updates the link correctly.

